PR #2, opened 2024-01-11
Feat/activation checkpointing

le1nux:


src/modalities/activation_checkpointing.py
```diff
@@ -0,0 +1,24 @@
...
+    apply_activation_checkpointing,
+    checkpoint_wrapper,
+)
+from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
+
+from modalities.models.gpt2.gpt2_model import Block
+
+
+def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module):
+    return isinstance(submodule, Block)
```
le1nux, line 15: `Block` is part of the GPT2 Model and hardcoded here. I think it would be better to have this configurable via the config, in case we have different blocks for other models in the future.  

We could think of having a `Blocks` enum in the config and parameterise this via config yaml.

le1nux (commented):
Implementation looks good to me functionality-wise. 👍 Left two comments regarding the integration in modalities which we should address.

config_files/config.yaml
```diff
@@ -111,16 +111,19 @@ training:
...
     type_hint: SaveKMostRecentCheckpointsStrategy
     config:
-      k: -1   # -1 to save all checkpoints
+      k: -1 # -1 to save all checkpoints
   checkpointing_execution:
     type_hint: FSDPToDiscCheckpointing
     config: 
       checkpoint_path: /path/to/checkpoints
+    config:
+      checkpoint_path: checkpoints
```
le1nux, line 126: This is a duplicate to the two lines above.

config_files/config.yaml
```diff
@@ -111,16 +111,19 @@ training:
   main_rank: 0
   train_batch_size: ${data.train_dataloader.config.batch_sampler.config.batch_size}
   global_num_seen_samples: ${modalities_setup.settings.global_num_seen_samples}
+  do_apply_activation_checkpointing: True
```
le1nux, line 114: We could think about adding `ActivationCheckpointing` a component. 
See my comment below regarding introducing a `Block` enum.

src/modalities/__main__.py
```diff
@@ -205,6 +206,10 @@ def construct_components(
         wrapped_model, optimizer = self.get_model_and_optimizer(
             config=config, running_env=running_env, checkpointing=checkpointing
         )
+        if config.training.do_apply_activation_checkpointing:
+            apply_activation_checkpointing_inplace(wrapped_model)
+            logging.info("Applied activation checkpointing!")
```
le1nux, line 211: Do we need this logging information? If so, I would say "enabled", rather than "applied".

PR #8, opened 2024-01-15
added support for using multiple losses and metrics in evaluator

BlueCrescent:


tests/test_evaluator.py
```diff
@@ -1,41 +1,100 @@
...
         local_rank=int(os.getenv("LOCAL_RANK")),
         batch_progress_publisher=progress_publisher_mock,
         evaluation_result_publisher=progress_publisher_mock,
     )
 
+    other_loss_mock = MagicMock(spec=Loss, return_value=torch.rand(1, requires_grad=True))
+
     evaluator.evaluate(
         model=nn_model_mock,
         data_loaders=[llm_data_loader_mock],
```
rrutmann, line 50: Might make sense to add a test for multiple data_loaders

src/modalities/gym.py
```diff
@@ -66,7 +66,7 @@ def _run_evaluation_and_checkpointing(
         eval_result = self.evaluator.evaluate(
             model=model,
             data_loaders=evaluation_data_loaders,
-            loss_fun=self.loss_fun,
+            loss_functions=[self.loss_fun],
```
rrutmann, line 69: Here we assume only one loss function. Where are multiple losses handled?

le1nux (changes requested):
I like the structure of the code and the introduction of small functions. Helps reading the code a lot! 

I was thinking about the next steps for the metrics and losses. They should be configurable by the config and assembled dynamically.

le1nux:
Is this PR finalised and ready for review?

tests/test_evaluator.py
```diff
@@ -1,41 +1,100 @@
...
-        loss_fun=loss_mock,
+        loss_functions=[loss_mock, loss_mock, other_loss_mock],
+        metrics=[metric_mock],
         global_train_sample_id=0,
         local_sample_id_to_global_sample_id=lambda i: i,
     )
-    nn_model_mock.forward.assert_has_calls([call(b.samples) for b in batches])
+    assert loss_mock.call_count == 4 * 2
```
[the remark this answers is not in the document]
BlueCrescent: Resolved in current version.

tests/test_evaluator.py
```diff
@@ -1,41 +1,100 @@
...
+
+
+def test_evaluate_calls_all_metrics(
```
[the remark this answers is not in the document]
BlueCrescent: Resolved in current version.

src/modalities/loss_functions.py
```diff
@@ -24,11 +24,11 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
 
 
 class CLMCrossEntropyLoss(Loss):
-    def __init__(self, target_key: str, prediction_key: str, tag: str = "CLMCrossEntropyLoss"):
+    def __init__(self, target_key: str, prediction_key: str, tag: str = "CLMCrossEntropyLoss", reduction: str = "sum"):
```
BlueCrescent, line 27: I think, we should set this reduction to mean. Even, if that means that the loss aggregation for logging purposes gets a bit wonky.
Suddenly switching to sum here will have unforeseen effects on model training. This is probably equivalent to multiplying the learning rate with batch_size * max_sequence_length.

src/modalities/evaluator.py
```diff
@@ -120,6 +77,51 @@ def evaluate(
...
+            )
+            global_dataset_sample_id = local_sample_id_to_global_sample_id(local_dataset_sample_id)
+            Evaluator._publish_progress(
+                batch_progress_publisher=self.batch_progress_publisher,
+                global_train_sample_id=global_train_sample_id,
+                global_dataset_sample_id=global_dataset_sample_id,
+                dataloader_tag=data_loader.dataloader_tag,
+            )
+
+        return {loss: loss.compute() for loss in losses}, {metric: metric.compute() for metric in metrics}
```
le1nux, line 109: maybe it makes sense to simplify losses and metrics just as measures? Maybe we can find a better word even?

le1nux: Also, what about replacing `compute` with `aggregate`?

src/modalities/evaluator.py
```diff
@@ -120,6 +77,51 @@ def evaluate(
...
         return result_dict
 
+    def _process_batches_in_data_loader(
+        self,
+        model: NNModel,
+        data_loader: LLMDataLoader,
+        global_train_sample_id: int,
+        local_sample_id_to_global_sample_id: Callable[[int], int],
+    ) -> Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]:
+        losses = [f.create(self.local_rank) for f in self._loss_factories]
```
le1nux, line 87: not a big fan of the factories here, as they seem a bit overengineered. 
How about we instantiate the aggregated measures already in the hierarchical instantiation and pass the measures to the evaluator in main?
We could implement a `reset` function for all measures that clears out their internal state. In this case, we could get rid of the factories here, right?

src/modalities/evaluator.py
```diff
@@ -147,3 +149,10 @@ def _publish_evaluation_result(
     @staticmethod
     def _get_local_sample_id(batch_id: int, batch_size: int) -> int:
         return (batch_id + 1) * batch_size - 1
+
+
+def _extract_num_samples(data_loader: LLMDataLoader) -> int:
+    num_samples = len(data_loader.dataset)
+    if data_loader.batch_size is not None and data_loader.drop_last:
```
le1nux, line 156: In `LLMDataLoader` the `batch_size` is never `None`.

src/modalities/evaluation/throughput_aggregator.py
```diff
@@ -0,0 +1,92 @@
...
+        a.start()
+
+
+class ThroughputAggregationContext:
+
+    def __init__(
+        self,
+        num_samples: int,
+        local_rank: int,
+        throughput_aggregator_factory: Callable[[], ThroughputAggregator] = ThroughputAggregator,
```
le1nux, line 70: Could we pass in the aggregator directly? The `reset` function is implmented already.

src/modalities/evaluation/perplexity.py
```diff
@@ -0,0 +1,54 @@
...
+    def _calc_measure(self, values: Dict[PerplexityKeys, torch.Tensor]) -> torch.Tensor:
+        return values[PerplexityKeys.PERPLEXITY] / values[PerplexityKeys.NUM_SAMPLES]
+
+
+class AggregativePerplexityFactory(AggregativeMeasureFactory[PerplexityKeys]):
+    def __init__(self, target_key: str, prediction_key: str) -> None:
+        self._target_key = target_key
+        self._prediction_key = prediction_key
+
+    def create(self, local_rank: int) -> AggregativeMeasure[PerplexityKeys]:
```
le1nux, line 49: as mentioned earlier, I think the factories are overkill.

le1nux (commented):
Really good work on the evaluation/aggregation part. I left a bunch of comments (some minor and some that would need some discussion). Primarily, I'm not so much convinced about the metric and loss factories that get instantiated on a very low level. I would  prefer if the instantiation would completely happen in the hierarchical instantiation and that these aggregated measures (instead of the factories) would be passed to the evaluator. I think that's the main point that necessitates discussion and I would like to hear your thoughts on it :-) 

FYI,  I did not take a look at the tests yet and wanted to add my comments already, as I think they would be helpful already.

src/modalities/trainer.py
```diff
@@ -64,18 +42,17 @@ def train(
...
         # batch loop
         batch: DatasetBatch
         # TODO: why do we need a barrier here?
         dist.barrier()
-        forward_backward_time_recorder = TimeRecorder()
-        forward_backward_time_recorder.start()
-        for batch_id, batch in enumerate(train_loader):
-            # Because we might resume training, we add the starting batch id of the data loader
+
+        for thoughput_aggregator, (batch_id, batch) in start_throughput_measurement(
```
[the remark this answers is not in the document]
BlueCrescent: I adapted this to use the context manager and not the weird iterable decorator anymore.

PR #9, opened 2024-01-15
CI/setup

lllAlexanderlll:
* CI for tests
* CI for linting
* Ci for build and deploy documentation (Wiki + API) to https://modalities.github.io/modalities/


config_files/config_lorem_ipsum.yaml
```diff
@@ -5,7 +5,7 @@ modalities_setup:
 
 wandb:
   project_name: modalities
-  mode: ONLINE
+  mode: OFFLINE
```
le1nux, line 8: Shouldn't the default be `ONLINE`? Otherwise, the user has to manually call the syncing script.

docs/source/configuration.rst
```diff
@@ -0,0 +1,89 @@
+.. role:: python(code)
```
le1nux, line 1: These files have the same content as the READMEs right? Is it possible to generate the rst files from the markdown? Otherwise, documentation always has to be updated in two places.

docs/source/index.rst
```diff
@@ -0,0 +1,53 @@
+Welcome to Modalities's documentation!
```
le1nux, line 1: Shouldn't it be "Modalities' documentation"? I don't want to be overly critical, but this is on the first page in the first line ... ;-)

tests/dataloader/test_packed_dataset.py
```diff
@@ -26,8 +25,7 @@ def test_packed_megatron_dataset_loading(dummy_packed_data_path, block_size, exp
 def test_packed_continuous_dataset_loading(dummy_packed_data_path, block_size, expected_length, expected_output):
     ds = PackedMemMapDatasetContinuous(dummy_packed_data_path, block_size, sample_key="input_ids")
     assert len(ds) == expected_length
-    dl = LLMDataLoader(dataloader_tag="unittest", dataset=ds)
-    retrieved_input_ids = [list(x["input_ids"]) for x in dl]
+    retrieved_input_ids = [list(batch["input_ids"]) for batch in ds]
```
le1nux, line 28: This change does not make sense to me. Why does `PackedMemMapDatasetContinuous` provide a `batch` in the for loop?

le1nux (changes requested):
Apart from my comments, this looks good to me! I mainly covered the code part. 
@luzian-hahn could you take a look at the GitHub workflows and docs generation, please?

tests/dataloader/test_packed_dataset.py
luzian-hahn: I guess you are right, it would be rather `packed_samples`. Is the other change clear (the removal of the data_loader?

.github/workflows/build_and_deploy_documentation.yml
```diff
@@ -0,0 +1,36 @@
...
+        pip install .
+        pip install sphinx-rtd-theme sphinx-autodoc-typehints sphinx-click sphinx-automodapi texext
+    - name: "Parse into HTML"
+      run: |
+        sphinx-apidoc -o docs/source/api src/modalities
+        sphinx-build -M html docs/source/ docs/build/
+    - name: Deploy to GitHub Pages
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        publish_branch: gh-pages
```
luzian-hahn, line 30: Does this mean, that the content is published to this branch? 
If so, isn't there the chance for conflicts when storing the documention on  parallel branch to main?

.github/workflows/build_and_deploy_documentation.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main 
+      - ci/setup
+
+permissions:
+  contents: write
```
luzian-hahn, line 11: Why is this necessary? Because we are pushing content to the branch "gh-pages"? If I understood this setting correctly, it indicates the permission to create "git-objects"

.github/workflows/build_and_deploy_documentation.yml
```diff
@@ -0,0 +1,36 @@
+name: "Build Sphinx Documentation"
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main 
+      - ci/setup
```
luzian-hahn, line 8: before merging, this should get removed

.github/workflows/build_and_deploy_documentation.yml
luzian-hahn: okay I found your comment about this, that the `gh-pages` branch "listens" to `main`. What do you mean by that?

config_files/config_lorem_ipsum.yaml
luzian-hahn: I am not sure what this setting does. However we keep on encountering issues with WandB if the tests (this config is only for the unittests) run on different devices, which is always based on the fact that not everybody uses WandB (especially the CI test setup). Therefore deactivating this in the tests makes much sense in my opinion. 
I deactivated them beforehand by using a different subscriber, but if this here is also an option, I do very much appreciate it.

luzian-hahn:
I fixed the tests for a CPU-only environment and[ tested them once again here in this PR](https://github.com/Modalities/modalities/actions/runs/7567511616/job/20606816686?pr=9).
Unfortunately I was only able to do so, by deactivating GPU-only features within the lorem_ipsum test scenario. If we want our CI-unittests to verify a working setup with GPUs, we might want to add another runner with GPUs to have the tests executed on it.

config_files/config_lorem_ipsum.yaml
le1nux: This setting stores the WandB logs only locally and does not sync them automatically. The user has to call a WandB syncing routine to push the logs to the WandB server. 

If this config is solely used for testing, then we can leave it as is. At the same time, we should have another config, which logs to WandB automatically to avoid confusion for the users.

tests/dataloader/test_packed_dataset.py
le1nux: yes the other changes are clear to me. I assume the intention behind them was to have the testing more granular? 
`packed_samples` would indeed be a better naming.

docs/source/configuration.rst
lllAlexanderlll: - discussed in the group: Have the Wiki for tutorials separate. The readme should mainly contain our USPs.

.github/workflows/build_and_deploy_documentation.yml
lllAlexanderlll: The Github action workflow listens to main and pushes the HTML files to the gh-pages branch.

lllAlexanderlll: We store the raw rst files on main, but not the HTML files. These will be created by sphinx and pushed only to the gh-pages branch

.github/workflows/build_and_deploy_documentation.yml
lllAlexanderlll: Yes, because we want to create/write files in gh-pages

docs/source/index.rst
lllAlexanderlll: Changed :)

.github/workflows/build_and_deploy_documentation.yml
lllAlexanderlll, line 8:
```suggestion
```

lllAlexanderlll: Done

le1nux:
LGTM :) Thanks for addressing the remarks.

PR #16, opened 2024-01-22
Add Contrastive Captioner (CoCa)

spravil:
This PR has the following updates:
- [X] Adds `einops` as dependency
- [x] Creates the `nn` subpackage for reuseable modules
- [X] Adds attention with cross-attention and non-causal sequence modeling
- [X] Adds the vision transformer (https://arxiv.org/abs/2010.11929)
- [X] Adds coca (https://arxiv.org/abs/2205.01917)
- [x] Adds attention pooling
- [X] Add test for coca forward function
- [x] Add full integration test (with dataloading mock)


src/modalities/models/vision_transformer/vision_transformer_model.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__()
+        self.conv = nn.Conv2d(n_img_channels, n_embd, kernel_size=patch_size, stride=patch_stride)
+        self.cls_token = None
+        if add_cls_token:
+            self.cls_token = nn.Parameter(torch.zeros(1, 1, n_embd))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        B, *_ = x.shape
+        x = self.conv(x)
+        x = rearrange(x, "b c h w -> b (h w) c")
```
spravil, line 46: Use Rearrange layer instead. See https://github.com/arogozhnikov/einops/wiki/Using-torch.compile-with-einops

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,193 @@
...
+        input_ids = inputs[self.sample_key]
+        device = input_ids.device
+        b, t = input_ids.size()
+        assert (
+            t <= self.config.block_size
+        ), f"Cannot forward sequence of length {t}, block size is only {self.config.block_size}"
+        pos = torch.arange(0, t + 1, dtype=torch.long, device=device)
+
+        tok_emb = self.transformer.wte(input_ids)
+        tok_emb = torch.cat([self.cls_token.repeat(b, 1, 1), tok_emb], dim=1)
```
spravil, line 62: Append cls token

src/modalities/models/coca/coca_model.py
```diff
@@ -68,17 +83,26 @@ def forward(self, inputs: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
...
-        if config.activation == ActivationType.GELU:
-            self.mlp = TransformerMLP(config)
-        elif config.activation == ActivationType.FUSED_SWIGLU:
-            hidden_dim = 256 * ((int(2 * 4 * config.n_embd / 3) + 256 - 1) // 256)
-            self.mlp = xops.SwiGLU(config.n_embd, hidden_dim, config.n_embd, bias=False)
+        self.ln_1 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+        self.attn = Attention(n_embd, n_head, bias, is_causal=True, use_cross_attention=True)
+        self.ln_2 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+
+        if activation == ActivationType.GELU:
```
thomaschhh, line 101: @mali-git for the sake of avoiding code duplication, would it helpful to somehow extract the selection of the activation or even the entire block as a whole? I am asking because the implementation is almost identical to the one in the [GPT2 Model](https://github.com/Modalities/modalities/blob/feat/coca/src/modalities/models/gpt2/gpt2_model.py#L196) except for the attention layer.

config_files/config_example_coca.yaml
```diff
@@ -0,0 +1,206 @@
...
+          text_sample_key: ${data.sample_key}
+          text_target_key: ${data.target_key}
+  eval_dataloaders:
+    - type_hint: LLMDataLoader
+      config:
+        num_workers: 2
+        pin_memory: true
+        shuffle: false
+        dataloader_tag: "val"
+        dataset: ${data.train_dataloader.config.dataset}
```
BlueCrescent, line 59: Linking the train data in val and test might encourage bad practices (even if it's just a dummy dataset right now).

config_files/config_example_coca.yaml
```diff
@@ -0,0 +1,206 @@
...
+                num_replicas: ${training.world_size}
+                shuffle: true
+        collate_fn: ${data.train_dataloader.config.collate_fn}
+    - type_hint: LLMDataLoader
+      config:
+        num_workers: 2
+        pin_memory: true
+        shuffle: false
+        dataloader_tag: "test"
+        dataset: ${data.train_dataloader.config.dataset}
```
BlueCrescent, line 78: Linking the train data in val and test might encourage bad practices (even if it's just a dummy dataset right now).

src/modalities/nn/mlp.py
```diff
@@ -0,0 +1,31 @@
...
+
+class MLP(nn.Module):
+    def __init__(
+        self,
+        in_features: int,
+        hidden_features: Optional[int] = None,
+        out_features: Optional[int] = None,
+        bias: bool = True,
+        dropout: float = 0.0,
+        act_fn: nn.Module = nn.GELU,
```
BlueCrescent, line 14: Better type hint would be Callable[[], nn.Module] to allow parameterized activation functions, e.g. "lambda: nn.LeakyReLU(negative_slope=0.02)".

src/modalities/models/coca/collator.py
```diff
@@ -0,0 +1,30 @@
...
+import torch
+
+from modalities.batch import DatasetBatch
+
+
+class CoCaCollator:
+    def __init__(self, sample_keys: List[str], target_keys: List[str], text_sample_key: str, text_target_key: str):
+        self.device: torch.device = field(default_factory=lambda: torch.device("cpu"))
+        assert text_sample_key in sample_keys, f"{text_sample_key} is not part of sample keys {sample_keys}"
+        assert text_target_key not in target_keys, f"{text_target_key} should not be part of target keys {target_keys}"
```
BlueCrescent, line 13: Better to raise ValueErrors.

src/modalities/models/coca/coca_model.py
BlueCrescent, line 1: Might be worth to consider splitting this file into multiple files for each major module.

BlueCrescent (approved):
This PR looks good. I added some comments, but none of them are critical. So I approved the PR.

However, please consider adding tests for all the Modules added (also in vision_transformer and nn). At least some shape tests that clearly document the expected input and output shapes.

tests/models/coca/test_coca.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_attn_pool():
+    model = AttentionalPooling(n_embd=768, n_head=8, bias=False, epsilon=1e-5)
+    dummy_vision_embed = torch.randn(1, 256, 768)
+    dummy_queries = torch.randn(1, 257, 768)
+    out = model(dummy_vision_embed, dummy_queries)
+    assert out.shape == (1, 257, 768)
+
+
```
flxst, line 51: You might want to use a `@pytest.mark.skipif` decorator in order to skip the end-to-end test if no GPU is available. See `modalities/tests/test_main.py`. https://github.com/Modalities/modalities/blob/d71bceb661b560ee9ccaa2a1ed25a089947d095c/tests/test_main.py#L11

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+from modalities.models.vision_transformer.vision_transformer_model import VisionTransformer, VisionTransformerConfig
+from modalities.nn.attention import Attention
+
+
+class TextDecoder(NNModel):
+    def __init__(self, config: GPT2Config):
+        super().__init__()
+        self.config = config
+        assert config.vocab_size is not None
+        assert config.block_size is not None
```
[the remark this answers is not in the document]
spravil: I removed these asserts as they should already be handled by pydantic

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+        x = x + self.mlp(self.ln_2(x))
+        return x
+
+
+class MultiModalDecoder(NNModel):
+    def __init__(self, config: GPT2Config):
+        super().__init__()
+        self.config = config
+        assert config.vocab_size is not None
+        assert config.block_size is not None
```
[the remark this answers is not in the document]
spravil: I removed these asserts as they should already be handled by pydantic

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+        super().__init__()
+        self.ln_1 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+        self.attn = Attention(n_embd, n_head, bias, is_causal=False, use_cross_attention=True)
+        self.ln_2 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+
+        if activation == ActivationType.GELU:
+            self.mlp = TransformerMLP(n_embd=n_embd, ffn_hidden=ffn_hidden, bias=bias, dropout=dropout)
+        elif activation == ActivationType.FUSED_SWIGLU:
+            hidden_dim = 256 * ((int(2 * 4 * n_embd / 3) + 256 - 1) // 256)
+            self.mlp = xops.SwiGLU(n_embd, hidden_dim, n_embd, bias=False)
```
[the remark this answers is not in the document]
spravil: This calculation was taken from our GPT2 implementation. It seems like a relict and i removed this line and use fft_hidden instead.

src/modalities/models/coca/collator.py
spravil: Done

src/modalities/models/coca/coca_model.py
spravil: I created separate files for the text decoder and the multimodal decoder

config_files/config_example_coca.yaml
```diff
@@ -0,0 +1,206 @@
...
+                rank: ${training.global_rank}
+                num_replicas: ${training.world_size}
+                shuffle: true
+        collate_fn: ${data.train_dataloader.config.collate_fn}
+    - type_hint: LLMDataLoader
+      config:
+        num_workers: 2
+        pin_memory: true
+        shuffle: false
+        dataloader_tag: "test"
```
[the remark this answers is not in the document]
spravil: I removed the dummy test data from the config

config_files/config_example_coca.yaml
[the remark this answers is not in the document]
spravil: The val dataloader now has its own dataset instance

config_files/config_example_coca.yaml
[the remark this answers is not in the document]
spravil: The val dataloader now has its own dataset instance

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+                    p, mean=config.weight_init.mean, std=config.weight_init.std / math.sqrt(2 * config.n_layer)
+                )
+
+    def _init_weights(self, module: nn.Module):
+        if isinstance(module, nn.Linear):
+            torch.nn.init.normal_(module.weight, mean=self.config.weight_init.mean, std=self.config.weight_init.std)
+            if module.bias is not None:
+                torch.nn.init.zeros_(module.bias)
+        elif isinstance(module, nn.Embedding):
+            torch.nn.init.normal_(module.weight, mean=self.config.weight_init.mean, std=self.config.weight_init.std)
```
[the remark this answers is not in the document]
spravil: I moved this weight init code to the main coca model.

tests/models/coca/test_coca.py
spravil: Good point, i've added the deorator to the e2e test

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,90 @@
...
+        epsilon: float,
+        activation: ActivationType,
+        n_head: int,
+        dropout: float,
+        ffn_hidden: int,
+    ):
+        super().__init__()
+        self.ln_1 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+        self.attn = Attention(n_embd, n_head, bias, is_causal=False, use_cross_attention=True)
+        self.ln_2 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
```
flxst, line 26: Don't we need a self-attention layer with causal masking first, before the cross-attention layer?

flxst (changes requested):
This PR looks great! I added a few comments. Most of them are about minor issues and optional changes, but I also found 1 or 2 things that are potentially more serious.

src/modalities/nn/attention.py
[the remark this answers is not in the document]
spravil: I created a new issue (#63) to remove the duplicated code in the future. Currently the aim is to leave the GPT2 implementation untouched.

src/modalities/nn/attention_pooling.py
```diff
@@ -0,0 +1,25 @@
...
+        n_head: int,
+        bias: bool,
+        epsilon: float,
+    ):
+        super().__init__()
+        self.ln_1 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+        self.attn = Attention(n_embd, n_head, use_cross_attention=True)
+        self.ln_2 = LayerNorm(ndim=n_embd, bias=bias, epsilon=epsilon)
+
+    def forward(self, vision_embd: torch.Tensor, vision_queries: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
spravil: I also changed the names to `queries` and `context`.

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if activation == ActivationType.GELU:
+            self.mlp = TransformerMLP(n_embd=n_embd, ffn_hidden=ffn_hidden, bias=bias, dropout=dropout)
+        elif activation == ActivationType.FUSED_SWIGLU:
+            self.mlp = xops.SwiGLU(n_embd, ffn_hidden, n_embd, bias=False)
+        else:
+            raise Exception("unimplemented activation")
```
le1nux, line 33:
```suggestion
            raise NotImplementedError(f"activation type {activation} not implemented")
```

tests/models/vision_transformer/vision_transformer_config.yaml
```diff
@@ -0,0 +1,13 @@
+sample_key: images
+prediction_key: logits
+block_size: 197
```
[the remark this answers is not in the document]
spravil: I added a function to calculate the block size based on the image size. The function is also tested for different input sizes, patch sizes and strides.

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,142 @@
...
+            self.prediction_key: logits,
+            self.vision_cls_prediciton_key: vision_cls_token,
+            self.text_cls_prediciton_key: text_cls_token,
+        }
+
+    def _forward_encode_vision(self, inputs):
+        vision_embd = self.vision_encoder(inputs)[self.vision_embd_prediciton_key]
+        queries = repeat(self.vision_queries, "n d -> b n d", b=vision_embd.shape[0])
+        vision_embd = self.attn_pool(vision_embd, queries)
+        vision_cls_token, vision_embd = vision_embd[:, :1, :], vision_embd[:, 1:, :]
```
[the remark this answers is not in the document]
le1nux: +1 this also confused me

le1nux (commented):
Very nice work! :)  Left a few comments.

src/modalities/models/coca/coca_model.py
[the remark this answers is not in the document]
spravil: Appending instead of prepending the cls token is common for vision and comes from the original vision transformer paper. I think it emphasizes that non-causal modeling is required, but has no other meaning. So I will prepend the cls token to avoid confusion.

tests/models/coca/test_coca.py
```diff
@@ -0,0 +1,65 @@
...
+    assert "text_cls" in out
+    assert out["logits"].shape == (1, 1024, 50304)
+    assert out["vision_cls"].shape == (1, 1, 768)
+    assert out["text_cls"].shape == (1, 1, 768)
+
+
+@pytest.mark.skipif(
+    no_gpu_available(), reason="This e2e test verifies a GPU-Setup and uses components, which do not support CPU-only."
+)
+def test_e2e_coca_training_run_without_checkpoint(monkeypatch):
```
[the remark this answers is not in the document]
spravil: For me the test works even with `fused_swiglu`. Maybe there is an issue with your xformers installation?

src/modalities/models/coca/multi_modal_decoder.py
spravil: For [open CoCa](https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/transformer.py#L188) they also only used Cross-Attention. However, an additional self-attention layer will probably help with loading pretrained LLM.

src/modalities/models/coca/text_decoder.py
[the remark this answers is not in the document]
spravil: I think it is likely that we will make changes to the text decoder in the future. So I would not couple it to gpt2 and keep the implementations separate, at least for now.

src/modalities/nn/attention_pooling.py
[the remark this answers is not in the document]
spravil: The attention pooling layer is currently only used by CoCa to aggregate information from the multimodal encoder outputs. I could move it to the CoCa package, but I think it would not hurt as a reusable module in the nn package.

src/modalities/nn/attention.py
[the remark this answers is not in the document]
spravil: There is an issue to address code duplication between CoCa and GPT2 #63.

src/modalities/models/coca/multi_modal_decoder.py
flxst: Had a brief look, I think they actually also use the self-attention layer (called `resblock`), see [here](https://github.com/mlfoundations/open_clip/blob/73fa7f03a33da53653f61841eb6d69aef161e521/src/open_clip/transformer.py#L790)

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+        self.lm_head = nn.Linear(in_features=config.n_embd, out_features=config.vocab_size, bias=False)
+
+        # init all weights
+        self.apply(self._init_weights)
+        # apply special scaled init to the residual projections, per GPT-2 paper
+        for pn, p in self.named_parameters():
+            if pn.endswith("c_proj.weight"):
+                torch.nn.init.normal_(
+                    p, mean=config.weight_init.mean, std=config.weight_init.std / math.sqrt(2 * config.n_layer)
+                )
```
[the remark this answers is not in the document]
spravil: Weight init is now moved to the main model class CoCa

src/modalities/dataloader/dataset.py
```diff
@@ -25,6 +25,36 @@ def _check_if_inbounds(self, idx: int):
...
+        return self.num_samples
+
+    def __getitem__(self, idx: int) -> Dict:
+        return self._create_random_sample()
+
+    def _create_random_sample(self):
+        sample = dict()
+        for sample_key, sample_shape, sample_type in self.sample_definition:
+            if sample_type == "float":
+                data = np.random.randn(*sample_shape)
```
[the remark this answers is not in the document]
spravil: I would rather add a global seed to the main configuration that is also applied to the weight init. For deterministic datasets we would need to cache the created tensors. If this is of interest, we should create an separate issue.

src/modalities/models/coca/text_decoder.py
```diff
@@ -0,0 +1,70 @@
...
+                        for _ in range(n_layer)
+                    ]
+                ),
+            )
+        )
+
+    def forward(self, inputs: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        input_ids = inputs[self.sample_key]
+        device = input_ids.device
+        b, t = input_ids.size()
```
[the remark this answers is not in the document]
spravil: Renamed to `B, T` as in the attention class.

src/modalities/nn/attention_pooling.py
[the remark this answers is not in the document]
spravil: I moved attention pooling to the coca subpackage

src/modalities/nn/attention.py
```diff
@@ -0,0 +1,77 @@
+import math
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import Tensor, nn
+
+
+class Attention(nn.Module):
```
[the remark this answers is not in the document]
spravil: I added an `AttentionConfig`, `AttentionType` and `AttentionEngineType`. `AttentionEngineType` is for selecting flash or default attention and can be set with the `AttentionConfig`. `AttentionType` is for causal, non-causal and cross attention and is usually hardcoded and cannot be changed by the user. Therefore it is a direct argument of `MultiHeadAttention`.

src/modalities/models/vision_transformer/vision_transformer_model.py
```diff
@@ -0,0 +1,122 @@
...
+    bias: bool = True
+
+
+class ImagePatchEmbedding(nn.Module):
+    def __init__(
+        self,
+        n_img_channels: int = 3,
+        n_embd: int = 768,
+        patch_size: int = 16,
+        patch_stride: int = 16,
```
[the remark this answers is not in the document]
spravil: For loss-less and non-overlapping patches `patch_stride = patch_size` is the correct setting. With `patch_stride = patch_size/2` you would have overlapping patches. Overlap can be beneficial, but is typically not used in Attention-based networks due to computational complexity. ConvNets, on the other hand, use overlap in all layers to build a large receptive field (for example ConvNeXt). Some works use Conv with overlap in the first layer and first blocks and apply Attention only in the later blocks (e.g. MetaFormer).

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+        self.multimodal_decoder = MultiModalDecoder(
+            sample_key=text_embd_prediciton_key,
+            prediction_key=text_decoder_config.prediction_key,
+            block_size=text_decoder_config.block_size,
+            n_layer=n_shared_layer,
+            **shared_decoder_kwargs,
+        )
+
+        # TODO Validate if weight tying is useful for coca
```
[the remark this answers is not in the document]
spravil: i created issue #69

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,142 @@
...
+        del shared_decoder_kwargs["block_size"]
+        del shared_decoder_kwargs["n_layer"]
+        del shared_decoder_kwargs["weight_init"]
+
+        self.text_decoder = TextDecoder(
+            sample_key=text_decoder_config.sample_key,
+            prediction_key=text_embd_prediciton_key,
+            block_size=text_decoder_config.block_size + 1,  # +1 for the class token
+            n_layer=text_decoder_config.n_layer - n_shared_layer,
+            **shared_decoder_kwargs,
```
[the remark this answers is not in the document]
spravil: Done!

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,90 @@
...
+        prediction_key: str,
+        block_size: int,
+        vocab_size: int,
+        n_layer: int,
+        n_head: int,
+        n_embd: int,
+        ffn_hidden: int,
+        dropout: float,
+        bias: bool,
+        attention: AttentionConfig,
```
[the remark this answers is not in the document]
spravil: Is now used to configure flash or default pytorch attention.

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,142 @@
...
+    def _forward_encode_vision(self, inputs):
+        vision_embd = self.vision_encoder(inputs)[self.vision_embd_prediciton_key]
+        queries = repeat(self.vision_queries, "n d -> b n d", b=vision_embd.shape[0])
+        vision_embd = self.attn_pool(vision_embd, queries)
+        vision_cls_token, vision_embd = vision_embd[:, :1, :], vision_embd[:, 1:, :]
+        return vision_embd, vision_cls_token
+
+    def _forward_encode_text(self, inputs):
+        text_embd = self.text_decoder(inputs)[self.text_embd_prediciton_key]
+        text_embd, text_cls_token = text_embd[:, :-1, :], text_embd[:, -1:, :]
```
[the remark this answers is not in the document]
spravil: This has no side effects because we first "build" the tuple, store it, and then override the variables. Side effects could only occure if we would use `text_embd` to index the output variables. See https://stackoverflow.com/questions/34171348/tuple-unpacking-order-changes-values-assigned

src/modalities/models/vision_transformer/vision_transformer_model.py
```diff
@@ -0,0 +1,122 @@
...
+        add_cls_token: bool = True,
+    ) -> None:
+        super().__init__()
+        self.conv = nn.Conv2d(n_img_channels, n_embd, kernel_size=patch_size, stride=patch_stride)
+        self.rearrange = Rearrange("b c h w -> b (h w) c")
```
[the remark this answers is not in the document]
spravil: Added the following lines for explaination:

```
# This is required to support torch.compile.
# See https://github.com/arogozhnikov/einops/wiki/Using-torch.compile-with-einops
```

src/modalities/models/vision_transformer/vision_transformer_model.py
```diff
@@ -0,0 +1,122 @@
...
+                for _ in range(n_layer)
+            ]
+        )
+
+        self.head = None
+        if n_classes is not None:
+            self.norm = nn.LayerNorm(n_embd)
+            self.head = nn.Linear(in_features=n_embd, out_features=n_classes, bias=bias)
+
+    def forward_embeddings(self, x: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
spravil: Renamed to `forward_images`. I want to keep the `forward_` prefix to emphasize that this is part of the forward pass.

src/modalities/nn/attention.py
```diff
@@ -0,0 +1,77 @@
...
+        y = self.resid_dropout(self.c_proj(y))
+        return y
+
+    def _forward_input_projection(self, x, context):
+        B, T, C = x.shape
+        _, Tc, Cc = context.shape
+        k = self.wk(context).view(B, Tc, self.n_head, Cc // self.n_head).transpose(1, 2)
+        q = self.wq(x).view(B, T, self.n_head, C // self.n_head).transpose(1, 2)
+        v = self.wv(context).view(B, Tc, self.n_head, Cc // self.n_head).transpose(1, 2)
+        return q, k, v
```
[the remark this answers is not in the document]
spravil: I added explainations for all variables. I think the code should be now sufficient to understand the output shapes.

src/modalities/nn/mlp.py
```diff
@@ -0,0 +1,31 @@
+from typing import Callable, Optional
+
+from torch import Tensor, nn
+
+
+class MLP(nn.Module):
```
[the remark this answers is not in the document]
spravil: I created issue #70 to address this. Different activation functions are already supported.

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,142 @@
...
+    text_decoder_config: GPT2Config
+    n_pool_head: Annotated[int, Field(ge=1)]
+    n_vision_queries: Annotated[int, Field(ge=1)]
+    bias_attn_pool: bool
+    epsilon_attn_pool: Annotated[float, Field(ge=0.0)]
+    n_shared_layer: Annotated[int, Field(ge=1)]
+
+
+class CoCa(NNModel):
+    """Contrastive Captioner"""
```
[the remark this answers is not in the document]
spravil: Done

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,176 @@
...
+            vocab_size=text_decoder_config.vocab_size,
+            n_head=text_decoder_config.n_head,
+            n_embd=text_decoder_config.n_embd,
+            ffn_hidden=text_decoder_config.ffn_hidden,
+            dropout=text_decoder_config.dropout,
+            bias=text_decoder_config.bias,
+            attention_config=text_decoder_config.attention_config,
+            activation=text_decoder_config.activation,
+            epsilon=text_decoder_config.epsilon,
+        )
```
flxst, line 99: It is a bit difficult to read and understand the input arguments for `TextDecoder`. It seems that most arguments are just passed forward from `text_decoder_config`, except `text_embd_prediction_key` (2nd argument), `block_size` (3rd argument) and `n_layer` (4th argument). We should perhaps consider passing the arguments as dictionaries (essentially `text_decoder_config` attributes with a few changed key-value pairs), similar to what is done for `VisionTransformer`. A concrete option might be to use something like
```
text_decoder_config_adjusted = {
    "prediction_key": text_embd_prediction_key, 
    "block_size": text_decoder_config.block_size + 1, 
    "n_layer": text_decoder_config.n_layer - n_shared_layer
}
text_decoder_config_adjusted = dict(text_decoder_config) | text_decoder_config_adjusted
self.text_decoder = TextDecoder(**text_decoder_config_adjusted)
```
which would make the changes to `text_decode_config` more explicit and everything a bit more readable in my opinion.

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,176 @@
...
```
flxst, line 114: same as above

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,176 @@
...
+        vision_cls_prediction_key: str,
+        text_cls_prediction_key: str,
+        vision_embd_prediction_key: str,
+        text_embd_prediction_key: str,
+        n_vision_queries: int,
+        n_pool_head: int,
+        bias_attn_pool: bool,
+        epsilon_attn_pool: float,
+        vision_encoder_config: VisionTransformerConfig,
+        text_decoder_config: GPT2Config,
```
flxst, line 73: I don’t like the name `text_decoder_config` very much, since it is used for both the `TextDecoder` and the `MultiModalDecoder`. Why not just `decoder_config`?

src/modalities/models/coca/coca_model.py
```diff
@@ -0,0 +1,176 @@
...
+    text_embd_prediction_key: str
+    vision_cls_prediction_key: str
+    text_cls_prediction_key: str
+    vision_encoder_config: VisionTransformerConfig
+    text_decoder_config: TextDecoderConfig
+    n_pool_head: Annotated[int, Field(ge=1)]
+    n_vision_queries: Annotated[int, Field(ge=1)]
+    bias_attn_pool: bool
+    epsilon_attn_pool: Annotated[float, Field(ge=0.0)]
+    n_shared_layer: Annotated[int, Field(ge=1)]
```
flxst, line 47: Is there a reason why `n_shared_layers` is not part of the `TextDecoderConfig`? One could have both `n_layer` and `n_shared_layer` as config attributes, or maybe even better, something like `n_layer_text` and `n_layer_multimodal` which would probably be simpler as it avoids having to deal with a subtraction (see line 89).

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,116 @@
...
+import xformers.ops as xops
+from torch import nn
+
+from modalities.models.gpt2.gpt2_model import ActivationType, LayerNorm
+from modalities.models.model import NNModel
+from modalities.nn.attention import AttentionConfig, AttentionType, MultiHeadAttention
+from modalities.nn.mlp import MLP
+
+
+class MultiModalBlock(nn.Module):
```
flxst, line 14: `MultiModalBlock` is also used for the pure `TextDecoder`. Hence, I think it would be better to call it `DecoderBlock`. In that case, it might also be good to remove the default values for the last two arguments `attention_type` and `with_context`, so that they always have to be explicitly specified no matter whether the `DecoderBlock` is used in the pure text or multimodal context.

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,116 @@
...
+        self.transformer = nn.ModuleDict(
+            dict(
+                h=nn.ModuleList(
+                    [
+                        MultiModalBlock(
+                            n_embd=n_embd,
+                            bias=bias,
+                            epsilon=epsilon,
+                            activation=activation,
+                            attention_config=attention_config,
```
flxst, line 97: `attention_config` should maybe be moved to last position, such that the order of arguments matches the one in `MultiModalBlock`'s `__init__`

src/modalities/models/vision_transformer/vision_transformer_model.py
```diff
@@ -0,0 +1,170 @@
...
+        if self.head:
+            if self.embedding_fn.cls_token is not None:
+                x = x[:, 0]
+            else:
+                x = x.mean(dim=1)
+            x = self.head(self.norm(x))
+        return {self.prediction_key: x}
+
+    @staticmethod
+    def _calcualte_block_size(img_size, patch_size, patch_stride, add_cls_token):
```
flxst, line 166: Typo in "calculate" and missing type hints. Also, would be nice to have a comment on where this formula is coming from, e.g. by adding a link.

src/modalities/nn/attention.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class AttentionEngineType(str, Enum):
+    DEFAULT_ATTENTION = "default_attention"
+    PYTORCH_FLASH_ATTENTION = "pytorch_flash_attention"
+
+
+class AttentionType(str, Enum):
+    CAUSAL_ATTENTION = "causal_attention"
+    NON_CAUSAL_ATTENTION = "non_causal_attention"
```
flxst, line 18: Maybe rename `causal_attention` to `causal_self_attention` and `non_causal_attention` to `non_causal_self_attention` in order to emphasize the distinction between self-attention and `cross_attention`?

flxst (commented):
Added a few comments and suggestions which are mostly about code quality / readability.

src/modalities/models/coca/coca_model.py
spravil: So this was basically the case before I addressed the following comment from @le1nux (See https://github.com/Modalities/modalities/pull/16/commits/d1f0369f4f19d07cb2a82d2e7717fe26126b00e5)
> instead of dumping key-value here I would use key-word arguments
> 
> ```python
> sample_key=shared_decoder_kwargs.sample_key
> prediction_key = shared_decoder_kwargs.prediction_key
> ...
> ```
> 
> The main reason is that if we `prediction_key` for instance the refactoring would not be reflected in the dictionary string keys.

I would prefer to keep the explicit keyword arguments for now.

src/modalities/models/coca/coca_model.py
spravil: I think `text_decoder_config` fits better because the `MultiModalDecoder` is also just a text decoder with additional cross-attention on multimodal input. Maybe it would be better to rename the `MultiModalDecoder` to `MultiModalTextDecoder`.

src/modalities/models/coca/coca_model.py
spravil: Sounds good, i will change that.

src/modalities/models/coca/multi_modal_decoder.py
spravil: Maybe the name `TransformerBlock` would be even better

src/modalities/models/coca/multi_modal_decoder.py
spravil: https://github.com/Modalities/modalities/pull/16/commits/be9e01d2156f7fc00f681286c9f202bb77afd5ec

src/modalities/models/coca/multi_modal_decoder.py
spravil: https://github.com/Modalities/modalities/pull/16/commits/be9e01d2156f7fc00f681286c9f202bb77afd5ec

src/modalities/models/coca/coca_model.py
spravil: https://github.com/Modalities/modalities/pull/16/commits/be9e01d2156f7fc00f681286c9f202bb77afd5ec

src/modalities/models/coca/coca_model.py
spravil: https://github.com/Modalities/modalities/pull/16/commits/be9e01d2156f7fc00f681286c9f202bb77afd5ec

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,123 @@
...
+        dropout: float,
+        ffn_hidden: int,
+        with_context: bool,
+        attention_type: AttentionType,
+        attention_config: AttentionConfig = None,
+        add_extra_mlp: bool = False,
+    ):
+        super().__init__()
+        self.with_context = with_context
+        self.add_extra_mlp = add_extra_mlp or not with_context
```
flxst, line 31:
```suggestion
        self.add_extra_mlp = add_extra_mlp
```

flxst: The `not with_context` part is redundant, as it also appears in lines 45 & 63. Alternatively, one could leave line 31 unchanged and reduce lines 45 & 63 to `if self.add_extra_mlp:`. However, the disadvantage of that approach is that we generally would have `self.add_extra_mlp != add_extra_mlp`, which might be confusing.

src/modalities/models/coca/multi_modal_decoder.py
```diff
@@ -0,0 +1,123 @@
...
+                            n_embd=n_embd,
+                            bias=bias,
+                            epsilon=epsilon,
+                            activation=activation,
+                            n_head=n_head,
+                            dropout=dropout,
+                            ffn_hidden=ffn_hidden,
+                            with_context=True,
+                            attention_type=AttentionType.CAUSAL_SELF_ATTENTION,
+                            attention_config=attention_config,
```
flxst, line 107:
```suggestion
                            attention_config=attention_config,
                            add_extra_mlp=False,
```

flxst: It might be useful to explicitly specify `add_extra_mlp` here (instead of relying on the default value), since `with_context` is `True` (line 105) which makes `add_extra_mlp` relevant.

flxst (commented):
Added 1-2 suggestions for improvement related to commit f415271cc70e122742146305e91108ae2c3dce7f

flxst (approved):

le1nux (approved):
LGTM :) Good work!

config_files/config.yaml
```diff
@@ -142,17 +141,13 @@ model:
     prediction_key: "logits"
     block_size: ${data.sequence_len}
     vocab_size: 50304 # GPT-2 vocab_size of 50257, padded up to nearest multiple of 64 for efficiency
-    n_layer: 12
-    n_head: 12
+    n_layer_q: 12
```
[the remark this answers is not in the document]
spravil: See #93

examples/getting_started/README.md
```diff
@@ -11,7 +11,7 @@ As a reference, this example has the following folder structure. Folders in <> w
     ├── example_config.yaml
     ├── data
     │   ├── mem_map
```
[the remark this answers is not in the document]
spravil: See #86

Issue #23, opened 2024-01-22
perf: benchmark dataloaders in comparison with MegatronLM and optimize if necessary


luzian-hahn:
## Motivation
Up to now the MegatronLM-implementation for Datasets was used (as is). Although relatively efficient, the downside is, that it was not developed by us and therefore in doubt difficult to maintain in the future.
Therefore we provided an own implementation. The main requirements are the usage of memmapped dataset inside the storage instead of the RAM. This should allow "indexed" access, while still not loading the whole dataset into memory. (One could also iterate over the dataset on disk, but this would not allow retrieving by any arbitrary index at any point in time).

## Goal
We want to benchmark both variants now and in case of nearly equal performance, drop the MegatronLM-native implementation.
This should happen by comparing three different runs with 
- 100b Tokens
- 250b Tokens
- 1t Tokens
respectively.

## Pitfalls
Right now MegatronLM's implementation uses more parallelism than we do, which will for sure result into much lower performance on our side. In order to have a fair comparison, one would need to do both preparation steps from our side (indexing, tokenized_indexing) and comparing it to the creation step of MegatronLM's implementation.

## Potential Optimizations
- Right now the indexing by us is using two different parallelizable tasks
  - loading chunks from the large file and putting it into a workload-queue. This requires a rather complex improvement to allow multiple threads reading from the same file, but still keeping things properly in order. One could e.g. do an initial binary search for newline chars and by this achieving nearly even distributed set of chunks, which can then be read via memory mapping or f.seek in parallel. Those can get processed in parallel as well, but the result needs to get concatenated in order.
  - the processing of the chunks is already threaded and can get easily extended without much effort. Only keeping the initial order would be something to verify
- The tokenized indexing can get parallelized since the usage of the previously created index allows parallel access to file content. However one needs to again ensure a sorted way of storing into the .pbin-file's date section. The tokenizer plays another major role regarding performance, so picking a fast option and giving efficient access per thread to it is also quite important.

PR #24, opened 2024-01-22
fix: support opengptx data encoding

luzian-hahn:
When playing around with the Dataset implementations of `PackedMemMapDataset` using OpenGPT-X data, decoding errors happened. This was also discovered in the past, that this data cannot get properly decoded using "utf8".
Although this very likely to be problem with the conversion of the respective data and the resulting encoding there, a potential solution is provided within this PR.

> **Note**: This branch is based on https://github.com/Modalities/modalities/pull/9 in order to use the respective test infrastructure and should get therefore merged afterwards, but not beforehand.

### DIFF
--- a/benchmarks/dataloader/README.md
+++ b/benchmarks/dataloader/README.md
@@ -1,0 +1,68 @@
+# Benchmarking of Dataset Implementations
+
+## Motivation
+We want to include a storage efficient, fast and generic dataset implementation in this repository.
+Previous work and ideas were based on MegatronLM and its dataset implementation.
+
+Unfortunately its usage is quite intransparent and causes regularly unexpected side effects.
+Those problems are hard to trace, as we are not the original authors of the code.
+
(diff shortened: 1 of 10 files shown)


luzian-hahn:
It is worth noting that this problem can also get solved by simplifying the general way of accessing files here. Right now we are basing the functionalities on `np.memmap`. We could however also use IOStreams with a certain offset and a specific `read()`-length. Some small attempts to play around with that indicate also much better performance (which makes it also interesting for #23. I decided however not to add it for now, since these changes might collide with the future use of more than text data.
If we want to add this anyhow, some more time needs to get invested to verify the stability of the code, since there seem to be still some problems with this implementation.

We can however also leave this to #23 and keep the changes here as a quick solution to support OpenGPT-X Data.

luzian-hahn:
Closing this, see https://github.com/Modalities/modalities/pull/40 for more details

PR #29, opened 2024-01-26
Add Noise Contrastive Estimation Loss

SogolHaghighat:
This PR contains the implementation and test for noise contrastive estimation loss for training multimodal models . Code adapted from (https://arxiv.org/pdf/1912.06430.pdf)

### DIFF
--- a/src/modalities/loss_functions.py
+++ b/src/modalities/loss_functions.py
@@ -46,8 +46,12 @@ class CLMCrossEntropyLoss(Loss):
 class NCELoss(torch.nn.Module):
     def __init__(self, asymm: bool, temp: float):
         """
-        if asymm is True the calculated loss will only have negative samples w.r.t. one modalityand.
-        If this value is False negative samples are w.r.t. both modalities. temp is the temprature value.
+        This calculates the noise contrastive estimation loss between embeddings of two different modalities
+        Implementation slightly adapted from https://arxiv.org/pdf/1912.06430.pdf by adding symmeric and asymmetric loss
+
+        Args:
+            asymm (bool): specifies if the loss is calculated in one direction or both directions
+            temp (float): temprature value for regulating loss
         """
         super(NCELoss, self).__init__()
         self.asymm = asymm
@@ -55,10 +59,13 @@ class NCELoss(torch.nn.Module):
 
     def forward(self, embedding1: torch.Tensor, embedding2: torch.Tensor, device: torch.device) -> torch.Tensor:
         """
-        Implementation slightly adapted from https://arxiv.org/pdf/1912.06430.pdf by adding symmeric and asymmetric loss
-        Calculates the noise contrastive estimation loss between embeddings of two different modalities
-        :input: batch of embedding1 and embedding2 of size bxd each, device of the embeddings
-        :return: Loss tensor
+        Args:
+            embedding1 (torch.Tensor): embeddings from modality 1 of size batch x dim
+            embedding2 (torch.Tensor): embeddings from modality 2 of size batch x dim
+            device (torch.device): torch device for calculating loss
+
+        Returns:
+            torch.Tensor: loss tensor
         """
         # calculating the similarity matrix of size (bxb)
         sim_matrix = torch.matmul(embedding1, embedding2.t()) / self.temp
@@ -76,12 +83,28 @@ class NCELoss(torch.nn.Module):
 
 class AsymmNCELoss(Loss):
     def __init__(self, prediction_key1: str, prediction_key2: str, tag: str = "AsymmNCELoss", temp: float = 1.0):
+        """
+        Asymmetric noise contrastive estimation loss
+
+        Args:
+            prediction_key1 (str): key to access embedding 1
+            prediction_key2 (str): key to access embedding 2
+            tag (str, optional): defaults to "AsymmNCELoss".
+            temp (float, optional): tempreture. Defaults to 1.0.
+        """
         super().__init__(tag)
         self.prediction_key1 = prediction_key1
         self.prediction_key2 = prediction_key2
         self.loss_fun = NCELoss(asymm=True, temp=temp)
 
     def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
+        """
+        Args:
+            forward_batch (InferenceResultBatch): data batch
+
+        Returns:
+            torch.Tensor: loss tensor
+        """
         embedding1 = forward_batch.get_predictions(self.prediction_key1)
         embedding2 = forward_batch.get_predictions(self.prediction_key2)
 
--- a/src/modalities/loss_functions.py
+++ b/src/modalities/loss_functions.py
@@ -43,74 +43,78 @@ class CLMCrossEntropyLoss(Loss):
         return loss
 
 
-class NCELoss(torch.nn.Module):
-    def __init__(self, is_asymmetric: bool, temperature: float):
+def nce_loss(
+    embedding1: torch.Tensor, embedding2: torch.Tensor, device: torch.device, is_asymmetric: bool, temperature: float
+) -> torch.Tensor:
+    """
+    This implementation calculates the noise contrastive estimation loss between embeddings of two different modalities
+    Implementation slightly adapted from https://arxiv.org/pdf/1912.06430.pdf, https://github.com/antoine77340/MIL-NCE_HowTo100M
+    changes include adding a temperature value and the choice of calculating asymmetric loss w.r.t. one modality
+
+    Args:
+        embedding1 (torch.Tensor): embeddings from modality 1 of size batch_size x embed_dim.
+        embedding2 (torch.Tensor): embeddings from modality 2 of size batch_size x embed_dim.
+        device (torch.device): torch device for calculating loss.
+        is_asymmetric (bool): boolean value to specify if the loss is calculated in one direction or both directions.
+        temperature (float): temperature value for regulating loss.
+
+    Returns:
+            torch.Tensor: loss tensor.
+    """
+    # calculating the similarity matrix of size (batch_size x batch_size)
+    sim_matrix = torch.matmul(embedding1, embedding2.t()) / temperature
+    # numerator of loss: using similarity scores for all positive pairs (e.g., image and its caption)
+    numerator = sim_matrix * torch.eye(sim_matrix.shape[0], device=device)
+    numerator = numerator.sum(dim=0).view(sim_matrix.shape[0], -1)
+    numerator = torch.logsumexp(numerator, dim=1)
+    # denominator of loss: using all similarity scores for all pairs (positive and negative)
+    if is_asymmetric:
+        denominator = torch.logsumexp(sim_matrix, dim=1)
+    else:
+        denominator = torch.logsumexp(torch.cat((sim_matrix, sim_matrix.t()), dim=1), dim=1)
+    return torch.mean(denominator - numerator)  # calculated in log space
+
+
+class NCELoss(Loss):
+    def __init__(
+        self,
+        prediction_key1: str,
+        prediction_key2: str,
+        is_asymmetric: bool = True,
+        temperature: float = 1.0,
+        tag: str = "NCELoss",
+    ):
         """
-        This calculates the noise contrastive estimation loss between embeddings of two different modalities
-        Implementation slightly adapted from https://arxiv.org/pdf/1912.06430.pdf, https://github.com/antoine77340/MIL-NCE_HowTo100M
-        changes include adding a temperature value and the choice of calculating asymmetric loss w.r.t. one modality
+        Noise Contrastive Estimation Loss
 
         Args:
-            is_asymmetric (bool): specifies if the loss is calculated in one direction or both directions
-            temperature (float): temperature value for regulating loss
-        """
-        super(NCELoss, self).__init__()
-        self.is_asymmetric = is_asymmetric
-        self.temperature = temperature
-
-    def forward(self, embedding1: torch.Tensor, embedding2: torch.Tensor, device: torch.device) -> torch.Tensor:
-        """
-        Args:
-            embedding1 (torch.Tensor): embeddings from modality 1 of size batch x dim
-            embedding2 (torch.Tensor): embeddings from modality 2 of size batch x dim
-            device (torch.device): torch device for calculating loss
-
-        Returns:
-            torch.Tensor: loss tensor
-        """
-        # calculating the similarity matrix of size (batch_size x batch_size)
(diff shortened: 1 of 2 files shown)


src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+            device (torch.device): torch device for calculating loss
+
+        Returns:
+            torch.Tensor: loss tensor
+        """
+        # calculating the similarity matrix of size (bxb)
+        sim_matrix = torch.matmul(embedding1, embedding2.t()) / self.temp
+        # numerator of loss: using similarity scores for all positive pairs (e.g., image and its caption)
+        numerator = sim_matrix * torch.eye(sim_matrix.shape[0], device=device)
+        numerator = numerator.sum(dim=0).view(sim_matrix.shape[0], -1)
```
le1nux, line 74: could this be simplified to `torch.diag(sim_matrix, 0)`?

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+        numerator = torch.logsumexp(numerator, dim=1)
+        # denominator of loss: using all similarity scores for all pairs (positive and negative)
+        if self.asymm:
+            denominator = torch.logsumexp(sim_matrix, dim=1)
+        else:
+            denominator = torch.logsumexp(torch.cat((sim_matrix, sim_matrix.permute(1, 0)), dim=1), dim=1)
+        return torch.mean(denominator - numerator)  # calculated in log space
```
le1nux, line 81: the paper uses `sum` instead of `mean`. 
Maybe `torch.sum(loss) / batch_size` makes more sense as we get the per sample loss?

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+        """
+        # calculating the similarity matrix of size (bxb)
+        sim_matrix = torch.matmul(embedding1, embedding2.t()) / self.temp
+        # numerator of loss: using similarity scores for all positive pairs (e.g., image and its caption)
+        numerator = sim_matrix * torch.eye(sim_matrix.shape[0], device=device)
+        numerator = numerator.sum(dim=0).view(sim_matrix.shape[0], -1)
+        numerator = torch.logsumexp(numerator, dim=1)
+        # denominator of loss: using all similarity scores for all pairs (positive and negative)
+        if self.asymm:
+            denominator = torch.logsumexp(sim_matrix, dim=1)
```
le1nux, line 78: I checked the paper and I am not sure if the two tests of positive and negative pairs, i.e., P_i and N_i, need to be of roughly equal size. Here, the positive examples would be largely underrepresented. Would this be an issue in practice?

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+        else:
+            denominator = torch.logsumexp(torch.cat((sim_matrix, sim_matrix.permute(1, 0)), dim=1), dim=1)
```
le1nux, line 80: is `sim_matrx.permute(1,0)` equivalent to transpose here? 
What is the effect here of making it "symmetric" (?) via torch.cat here?

tests/test_loss_functions.py
```diff
@@ -0,0 +1,21 @@
...
+    targets = {"target": torch.zeros(1024, 512)}
+    batch_dim = 1024
+    result_batch = InferenceResultBatch(targets, predictions, batch_dim)
+    return result_batch
+
+
+# calculating asymmetric NCELoss between a batch of embeddings and itself --> zero
+@pytest.mark.parametrize("key", ["embedding"])
+def test_AsymmNCELoss_is_zero(dummy_result_batch, key):
+    loss_func = AsymmNCELoss(key, key)
```
le1nux, line 20: maybe change to keyword arguments?

tests/test_loss_functions.py
```diff
@@ -0,0 +1,21 @@
...
+    predictions = {"embedding": torch.FloatTensor(1024, 512).uniform_(-50, 50)}
+    targets = {"target": torch.zeros(1024, 512)}
+    batch_dim = 1024
+    result_batch = InferenceResultBatch(targets, predictions, batch_dim)
+    return result_batch
+
+
+# calculating asymmetric NCELoss between a batch of embeddings and itself --> zero
+@pytest.mark.parametrize("key", ["embedding"])
+def test_AsymmNCELoss_is_zero(dummy_result_batch, key):
```
le1nux, line 19: what about adding another example for which we calculate the correct score manually and then we check this in a test case as well? Right now we only check for a batch of two identical embedding matrices.

le1nux (commented):
Thanks for your PR @SogolHaghighat 👍 

I left a few stylistic remarks and some questions. Once addressed we can merge :-)
 
Since I am not fully aware of the NCE loss function and just skimmed through the respective paper, I wanted to ask you if my understanding is correct. 

We start with a single batch that has two embedding matrices (e.g., one text matrix and one image matrix). The goal of the loss function is to learn similar for embeddings for correct (text, image) pairs and dissimilar embeddings for incorrect (text, image) pairs. Therefore we define the contrastive loss where we maximise the ratio of similarity(correct pairs) over similarity(all pairs). 
Is my understanding correct? Thanks :-)

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
         # Flatten the tokens
         loss = self.loss_fun(shift_logits.view(-1, shift_logits.size(-1)), shift_labels.view(-1))
         return loss
+
+
+class NCELoss(torch.nn.Module):
```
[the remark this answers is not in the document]
SogolHaghighat: NCELoss is not an existing loss function offered by torch. All losses implemented in torch inherit from _Loss which is a torch module (https://pytorch.org/docs/stable/_modules/torch/nn/modules/loss.html#L1Loss). By inheriting from torch.nn.Module after the calculation of the loss (forward pass) there will be a backward pass (loss.backward()) that calculates dloss/dx for every parameter x which has requires_grad=True. (https://discuss.pytorch.org/t/what-does-the-backward-function-do/9944). 
Loss interface (from what i understand) does not have a backward pass implemented by default. This causes problems in back-propagation when training. Looking at CLMCrossEntropyLoss we have in loss_functions.py, the actual loss function is CrossEntropyLoss() which is imported from torch.nn and by default should have the backward pass implemented.

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+    def __init__(self, asymm: bool, temp: float):
+        """
+        This calculates the noise contrastive estimation loss between embeddings of two different modalities
+        Implementation slightly adapted from https://arxiv.org/pdf/1912.06430.pdf by adding symmeric and asymmetric loss
```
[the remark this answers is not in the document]
SogolHaghighat: A better link might be this one: https://github.com/antoine77340/MIL-NCE_HowTo100M/blob/master/loss.py 
As can be seen, there is very minimal changes made to this code in comparison to NCELoss() here.
1. introducing the temperature, temperature is used in some other implementations therefore it is also added here for completeness and to have the option. by default it is assumed to be 1. 
2. asymm variable. asymm variable controls if the loss is only calculated with respect to one modality or in both directions. so lets say we have a batch of N paired samples of (image,text), for image1, only text1 is a positive match therefore for each image in the batch, there will be 1 positive pair and N-1 negative pairs. if the loss is calculated only using this sampling then it is asymmetric. however, in the original implementation, N-1 negative pairs w.r.t. each text is also considered in the bag of negative samples (a.k.a in the denominator of the loss). This is a bidirectional way of calculating loss or basically a symmetrical loss. 
It is very important to note two things: 
1. based on my understanding, the authors in their original implementation only take (image1, text1) once in the numerator as the positive sample. but in the denominator they take this twice (as the denominator contains all possible similarity scores between all possible pairs). this means the loss will not become zero even if two exact embeddings are used to calculate the loss. 
2. this paper follows an idea called multi-instance learning. this means potentially there can be more than one positive match for each image. For models such as CoCa this is not the case but for the sake of potential future use-cases, this has been kept this way. Also for CoCa model asymmetric loss is required.

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+
+        Args:
+            asymm (bool): specifies if the loss is calculated in one direction or both directions
+            temp (float): temprature value for regulating loss
+        """
+        super(NCELoss, self).__init__()
+        self.asymm = asymm
+        self.temp = temp
+
+    def forward(self, embedding1: torch.Tensor, embedding2: torch.Tensor, device: torch.device) -> torch.Tensor:
```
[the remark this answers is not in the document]
SogolHaghighat: As mentioned above, the costume losses inherit from nn.Module to allow for backward pass. by default these modules implement forward function. NCELoss is wrapped in the desired loss for CoCa (AsymmNCELoss) which inherits from Loss interface and should be callable.

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+            forward_batch (InferenceResultBatch): data batch
+
+        Returns:
+            torch.Tensor: loss tensor
+        """
+        embedding1 = forward_batch.get_predictions(self.prediction_key1)
+        embedding2 = forward_batch.get_predictions(self.prediction_key2)
+
+        shift_embedding1 = embedding1.contiguous()
+        shift_embedding2 = embedding2.contiguous()
```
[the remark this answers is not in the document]
SogolHaghighat: Maybe the naming is just confusing. The point is just to make sure tensors are contiguous to avoid any potential error.

src/modalities/loss_functions.py
```diff
@@ -41,3 +41,75 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+        Args:
+            prediction_key1 (str): key to access embedding 1
+            prediction_key2 (str): key to access embedding 2
+            tag (str, optional): defaults to "AsymmNCELoss".
+            temp (float, optional): tempreture. Defaults to 1.0.
+        """
+        super().__init__(tag)
+        self.prediction_key1 = prediction_key1
+        self.prediction_key2 = prediction_key2
+        self.loss_fun = NCELoss(asymm=True, temp=temp)
```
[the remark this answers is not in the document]
SogolHaghighat: So NCELoss is used as a costume loss and contains the main calculations of contrastive loss and AsymmNCELoss is just more like a wrapper which requires the instance of NCELoss and then inherits from Loss interface. I think they should be kept separate.

src/modalities/loss_functions.py
SogolHaghighat: `  numerator = sim_matrix * torch.eye(sim_matrix.shape[0], device=device)`
`  numerator = numerator.sum(dim=0).view(sim_matrix.shape[0], -1)`
`  numerator = torch.logsumexp(numerator, dim=1)`
So technically speaking, i think all above three lines could be replaced by `torch.diag(sim_matrix, 0)` however, I think for two reasons maybe authors wrote the code this way:

1. to keep the implementation close to the formula: 
formula

2. To allow for multiple instances of positive pairs per input. (MIL-NCE)

We could keep it as is and monitor if it causes any slow-down.

src/modalities/loss_functions.py
SogolHaghighat: I believe `torch.mean(denominator - numerator)` should take care of dividing sum of losses by the batch_size as both numerator and denominator have the dimensionality of batch_size

tests/test_loss_functions.py
SogolHaghighat: could you please clarify on this?

SogolHaghighat:
> Thanks for your PR @SogolHaghighat 👍
> 
> I left a few stylistic remarks and some questions. Once addressed we can merge :-)
> 
> Since I am not fully aware of the NCE loss function and just skimmed through the respective paper, I wanted to ask you if my understanding is correct.
> 
> We start with a single batch that has two embedding matrices (e.g., one text matrix and one image matrix). The goal of the loss function is to learn similar for embeddings for correct (text, image) pairs and dissimilar embeddings for incorrect (text, image) pairs. Therefore we define the contrastive loss where we maximise the ratio of similarity(correct pairs) over similarity(all pairs). Is my understanding correct? Thanks :-)

yes pretty much. the main difference in different papers is how they construct sets of positive and negative pairs.

src/modalities/loss_functions.py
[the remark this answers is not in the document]
le1nux: Thanks for the clarifications. I think I get the idea now :-)

le1nux (changes requested):
Thanks for the in-depth clarification, which helped my understanding of the intrinsics of the NCE loss a lot! Left a few minor suggestions. :-)

src/modalities/loss_functions.py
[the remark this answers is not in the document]
SogolHaghighat: since the loss does not need to be an `nn.Module`, after an internal discussion, the final change is that the calculation of contrastive loss is packed inside `nce_loss` function, and `NCELoss` now implements `Loss` interface where `is_asymmetric` can be configured in the constructor. For now `nce_loss` is kept inside `loss_functions.py` but if we implement more losses this way, maybe we can consider moving them into another location.

tests/test_loss_functions.py
SogolHaghighat: A new test is added to check the correctness of nce_loss function. 

Update to nce_loss:
as mentioned above the code was adapted from MIL-NCE loss from (https://arxiv.org/pdf/1912.06430.pdf). however, since the goal is to use this loss for training CoCa model, the code is now adapted to reflect the contrastive loss presented in CoCa paper (https://arxiv.org/pdf/2205.01917.pdf). There is still an optional variable for calculating the loss only w.r.t. one modality or in both directions. For CoCa model a bidirectional loss should be applied. 

Please could you review the changes and let me know of any other updates. Hand-written calculations of the loss examples are available in case you are interested. 

Many thanks for your review!

le1nux (approved):
Looks good to me! Merging :-)

Issue #28, opened 2024-01-26
Tests and linting actions are currently failing on main branch


spravil:
See [testing](https://github.com/Modalities/modalities/actions/runs/7615511938/job/20740206695) and [linting](https://github.com/Modalities/modalities/actions/runs/7615511935/job/20740206700) actions.

PR #38, opened 2024-01-30
fix: tests on cpu

flxst:
Attempts to fix the testing part of #28 (i.e. unit testing on CPU) by

- replacing `torch.device(self.local_rank)` by `torch.device(self.local_rank) if torch.cuda.is_available() else "cpu"`
- removing the mocking of the "reduce" operation which causes `test_gym.py` to fail

### DIFF
--- a/src/modalities/evaluator.py
+++ b/src/modalities/evaluator.py
@@ -45,11 +45,11 @@ class Evaluator:
     ) -> Dict[str, EvaluationResultBatch]:
         result_dict: Dict[str, EvaluationResultBatch] = {}
         model.eval()
+
+        device = torch.device(self.local_rank) if torch.cuda.is_available() else "cpu"
+
         for data_loader in data_loaders:
-            if torch.cuda.is_available():
-                cummulated_loss = torch.zeros(3).to(torch.device(self.local_rank))
-            else:
-                cummulated_loss = torch.zeros(3).to("cpu")
+            cummulated_loss = torch.zeros(3).to(device)
 
             Evaluator._publish_progress(
                 batch_progress_publisher=self.batch_progress_publisher,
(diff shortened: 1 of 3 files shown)


src/modalities/trainer.py
```diff
@@ -101,9 +103,7 @@ def train(
             # Check, if model should be evaluated
             if (local_train_batch_id + 1) % callback_interval_in_batches == 0:
                 if local_train_batch_id > 0:
-                    foward_backward_time = torch.tensor(forward_backward_time_recorder.delta_t).to(
-                        torch.device(self.local_rank)
-                    )
+                    foward_backward_time = torch.tensor(forward_backward_time_recorder.delta_t).to(device)
                     forward_backward_time_recorder.reset()
 
                     thoughput_aggregator.add_value(
```
flxst, line 109: The previously used mocking of the "reduce" operation led to problems below. More concretely, line 112 and line 115 both returned `None` (the `return_value`), which caused an error in line 118.

src/modalities/evaluator.py
```diff
@@ -45,11 +45,11 @@ def evaluate(
     ) -> Dict[str, EvaluationResultBatch]:
         result_dict: Dict[str, EvaluationResultBatch] = {}
         model.eval()
+
+        device = torch.device(self.local_rank) if torch.cuda.is_available() else "cpu"
```
flxst, line 49: This line also appears in `src/modalities/trainer.py` (line 69), we might want to use a function to avoid duplicate code.

src/modalities/trainer.py
```diff
@@ -66,6 +66,8 @@ def train(
         cummulated_loss = self._reset_loss()
         thoughput_aggregator = Aggregator[ThroughputAggregationKeys]()
 
+        device = torch.device(self.local_rank) if torch.cuda.is_available() else "cpu"
```
flxst, line 69: This line also appears in `src/modalities/evaluator.py` (line 49), we might want to use a function to avoid duplicate code.

flxst:
Note: linting is fixed in #42

src/modalities/evaluator.py
lllAlexanderlll: Fix code duplication later, as we usually run on GPU and only tests need CPU

lllAlexanderlll (approved):

lllAlexanderlll (approved):

le1nux (approved):

PR #39, opened 2024-01-30
Feat: RoPE Embeddings

mali-git:


src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -226,20 +361,38 @@ def __init__(
...
+            # wpe = nn.Embedding.from_pretrained(torch.zeros(block_size, n_embd))
+            wpe = nn.Identity()
+        else:
+            raise TypeError(f"{poe_type} not supported")
+
+        if poe_type is PositionTypes.ABSOLUTE and RotaryTransform in [
+            config.type_hint.value for config in attention.qkv_transforms
+        ]:
+            logging.warning("You are using RotaryTransform together with absolute position embeddings."
+                            " It is expected to use \"RotaryTransform\" together with \"NOPE.\"")
```
BlueCrescent, line 390: Should raise a ValueError. I don't see how having both could be useful.

Though, maybe we should add PositionTypes.ROTARY and in init add the RotaryTransform to qkv_transforms when that is set and it was not added yet.

tests/test_rotary_qkv_transform.py
```diff
@@ -0,0 +1,23 @@
...
+    rotary_transform = RotaryTransform(n_embd=embedding_dim, n_head=n_heads)
+
+    q_rot, k_rot, v_rot = rotary_transform(q=q, k=k,v=v)
+
+    assert not torch.equal(q, q_rot)
+    assert q.shape == q_rot.shape
+    assert not torch.equal(k, k_rot)
+    assert k.shape == k_rot.shape
+    assert torch.equal(v, v_rot)
+    assert v.shape == v_rot.shape
```
BlueCrescent, line 23: Ideally, create individual test for different asserts with descriptive test names.

src/modalities/models/gpt2/gpt2_model.py
BlueCrescent, line 1: This file is getting to long. We should split it up into multiple files, e.g. for the different modules.

Also, formatting is missing.

tests/test_rotary_qkv_transform.py
BlueCrescent, line 1: Formatting is missing for this file.

tests/test_evaluation.py
```diff
@@ -28,11 +29,12 @@ def test_pretrained_gpt_model(tmp_path):
...
         sample_key="input_ids",
         prediction_key="logits",
-        weight_init=WeightInitailizationConfig(mean=0, std=0.02),
+        weight_init=WeightInitializationConfig(mean=0, std=0.02),
+        poe_type=PositionTypes.NOPE
     )
 
     model = GPT2LLM(**pretrained_config)
-    model.save_pretrained(tmp_path)
+    #model.save_pretrained(tmp_path)
```
BlueCrescent, line 37: Does this test still work, if we don't save the model anymore?

Btw. this test gets moved/deleted by #34.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
...
+        "--gpus",
+        "all"
+    ],
+    "customizations": {
+        "vscode": {
+            "settings": {
+                "terminal.integrated.shell.linux": "/bin/bash"
+            },
+            "mounts": [
+                "source=/home/fhgiais/.bashrc,target=/root/.bashrc,type=bind"
```
BlueCrescent, line 16: This path looks pretty specific to our infrastructure. Does it make sense to check that in?

src/modalities/config/look_up_enum.py
BlueCrescent, line 1: Formatting might be missing.

src/modalities/config/utils.py
BlueCrescent, line 1: Formatting might be missing.

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -1,18 +1,116 @@
...
+
 
 # GPT2 implementation taken from nanogpt https://github.com/karpathy/nanoGPT
 
+class PositionTypes(str, Enum):
+    ABSOLUTE = "ABSOLUTE"
+    NOPE = "NOPE"
+
+
+class QueryKeyValueTransform(nn.Module):
```
BlueCrescent, line 27: It would make sense, to turn this into an abstract class with and abstract forward() method.

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -1,18 +1,116 @@
...
+
+        return self._cos_cached, self._sin_cached
+
+
+    def apply_rotary_pos_emb(self, x, cos, sin):
+        # NOTE: This could probably be moved to Triton
+
+        # Handle a possible sequence length mismatch in between q and k
+        cos = cos[:, :, : x.shape[-2], :]
+        sin = sin[:, :, : x.shape[-2], :]
```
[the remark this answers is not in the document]
mali-git: Done in 6d1bf285abfd3df692537cecfa2a3e5504e9b407.

src/modalities/models/gpt2/gpt2_model.py
mali-git: Done in 2ca59fb40d659cdb17b77275711eb4b1e45d34ab.

tests/test_rotary_qkv_transform.py
mali-git: We refactored the test in 9b474f3c0b08ee704bd74121a993d00b8f79b24c.

PR #42, opened 2024-01-30
fix: lint all files

flxst:
Fixes the linting part of #28.

As part of `git commit`, the pre-commit hooks only run on the **changed files**, corresponding to `pre-commit run`. However, as part of the github linting workflow, the pre-commit hooks are run on **all files**, corresponding to `pre-commit run --all-files`. Since the last few commits always included som (unchanged) files that did not adhere to the linting rules, the linting action failed.

I ran `pre-commit run --all-files` manually once and fixed the errors. That should fix the problem (at least for now).


flxst:
Note: tests are fixed in #38

PR #45, opened 2024-02-01
Hierarchical instantiation

le1nux:
Implements hierarchical instantiation of components via dependency injection. 

Given a config defined in a YAML file, we first load the config into a dictionary within the factory. We then traverse the dictionary recursively in depth-first fashion. Once we reach a leave component, we first instantiate the respective component config via Pydantic. Since the component config is self-contained, we leverage the referenced class type in the  type_hint to instantiate the component. We then return upwards to the parent item in the dictionary. If there are other child components in the parent component, then we also instantiate those. Once, all dependencies of the parent are instantiated, we can also instantiate the parent config and then the parent component. 

We traverse the entire config in this fashion by first instantiation the component configs and then the components themselves. Once all the components have been built, we return the root components whose keys we passed to the factory. 

This instantiation has a couple of advantages. 
1. we can define arbitrary configs and component dependencies, since pedantic is only applied to single components in isolation. Before we were limited by the `Appconfig` definition.
2. Registry is not necessary anymore. We only need to pass the `ConfigTypes` to the `Factory`.
3. Components in the config specify the dependent components with their actual types now. Before we had dependencies as config types , e.g., `DataloaderConfig` would have had a dependency for `DatasetConfig`. Now, the dependency can be either a concrete class e.g., `Dataset` or even more generic an Interface `DatasetIF`. In fact, the possibility for interface type checks makes sure that that the dependencies always implement the required interface and would throw an error otherwise. 

As of right now I don't see any disadvantage ... but would be great to discuss potential issues as well.


examples/getting_started/getting_started_example.md
flxst, line 1: Was this added by mistake?

src/modalities/config/component_factory.py
```diff
@@ -0,0 +1,129 @@
+from typing import Any, Dict, List, Union
+
+from pydantic import RootModel
+
+from tests.config.hierarchical_instantiation.configs import CompConfigABC
```
flxst, line 5: This should be `from tests.config.configs import CompConfigABC`. However, maybe we should consider moving the `CompConfigABC` class from `tests` to `modalities`, as it is not only used for testing but also here in the source code. This way, modules in `tests` will import from `modalities`, but not the other way around.

tests/config/configs.py
```diff
@@ -0,0 +1,89 @@
+from abc import ABC
+from enum import Enum
+from typing import Annotated, Any, List, Literal
+
+from pydantic import BaseModel, GetCoreSchemaHandler, validator
+from pydantic_core import core_schema
+
+from tests.config.hierarchical_instantiation.components import ComponentTypes, ComponentVWIF
```
flxst, line 8: This should be `from tests.config.components import ComponentTypes, ComponentVWIF`

tests/config/test_component_factory.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
+from typing import Dict, Union
+
+from omegaconf import OmegaConf
+
+from modalities.config.component_factory import ComponentFactory
+from tests.config.hierarchical_instantiation.configs import (
```
flxst, line 7: This should be `from tests.config.configs import (`

tests/config/test_component_factory.py
```diff
@@ -0,0 +1,32 @@
...
+)
+
+
+def load_app_config_dict(config_file_path: Path) -> Dict:
+    cfg = OmegaConf.load(config_file_path)
+    config_dict = OmegaConf.to_container(cfg, resolve=True)
+    return config_dict
+
+
+config_file_path = Path("config.yaml")
```
flxst, line 23: This relative path should be replaced by an absolute path. Otherwise, `pytest` will only work if executed from the file's directory, not the root directory. A solution could be to use `_ROOT_DIR` as in `tests/conftest.py`

tests/config/test_component_factory.py
```diff
@@ -0,0 +1,32 @@
...
+    CompYConfig,
+    CompZConfig,
+    ReferenceConfig,
+)
+
+
+def load_app_config_dict(config_file_path: Path) -> Dict:
+    cfg = OmegaConf.load(config_file_path)
+    config_dict = OmegaConf.to_container(cfg, resolve=True)
+    return config_dict
```
flxst, line 20: This function also exists in `src/modalities/__main__.py`. The only difference seems to be logging. If that is a problem, I would suggest to add an argument `verbose: bool = True` to the function, which skips the logging if `verbose` is `False`.

tests/config/test_component_factory.py
```diff
@@ -0,0 +1,32 @@
...
+config_file_path = Path("config.yaml")
+comp_config_types = Union[CompVConfig, CompWConfig, CompXConfig, CompYConfig, CompZConfig, ReferenceConfig]
+component_names = ["comp_z_1", "comp_x_1", "comp_y_1"]
+
+config_dict = load_app_config_dict(config_file_path=config_file_path)
+
+components = ComponentFactory.build_config(
+    config_dict=config_dict, config_types=comp_config_types, component_names=component_names
+)
+print(components)
```
flxst, line 32: Maybe instead of just checking that building the `components` does not throw an error, we should explicitly check that `components` has the dictionary structure and class instances we expect?

src/modalities/config/component_factory.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class ComponentFactory:
+    @staticmethod
+    def build_config(config_dict: Dict, config_types, component_names: List[str]) -> Dict[str, Any]:
+        component_dict = ComponentFactory._build_components(
+            config_dict=config_dict, config_types=config_types, component_names=component_names
+        )
+        return component_dict
```
flxst, line 14: It seems that `build_config` only calls another method, `_build_components`. Is there a reason to have 2 separate methods?

flxst (commented):
Nice work! I have left a few comments, they are only about code quality (so far).

tests/config/test_component_factory.py
le1nux: yes, the testing is not implmented yet

src/modalities/config/component_factory.py
le1nux: fixing it. Thanks!

tests/config/test_component_factory.py
le1nux: fixed. kept only the one in __main__.py

README.md
```diff
@@ -102,8 +103,8 @@ class SchedulerTypes(LookupEnum):
     ConstantLR = torch.optim.lr_scheduler.ConstantLR
 
 class StepLRConfig(BaseModel):
-    step_size: conint(ge=1)
-    gamma: confloat(ge=0.0)
+    step_size: Annotated[int, Field(strict=True, gt=1)]
+    gamma: Annotated[float, Field(strict=True, gt=0.0)]
```
le1nux, line 107: here, and also in other cases  `ge` was replaced by `gt`. Leaving it as a comment, that we have to fix it.

flxst: Oops, good catch! I fixed this in fb0aea5fe1fdb04dae9f6299ed2b5f39b4a3be17. Maybe we should consider using only `ge` for the sake of consistency. Type aliases might also help to avoid these kind of mistakes in the future, see  my comment here: https://github.com/Modalities/modalities/issues/57#issuecomment-1961687570

mali-git (approved):

fromm-m (approved):
approved

dertilo:
Hi modalities-people!
Can you help me understand the differences/advantages of your hierarchical instantiation approach compared to other open-source frameworks like [hydra](https://hydra.cc/docs/tutorials/structured_config/hierarchical_static_config/) or [pydantic](https://docs.pydantic.dev/latest/concepts/models/#nested-attributes)?

PR #67, opened 2024-03-04
RMS norm implementation

le1nux:
* Implemented RMS Norm based on https://arxiv.org/pdf/1910.07467.pdf and majorly adopted from https://github.com/facebookresearch/llama/blob/main/llama/model.py#L34C1-L77C36
* Refactored the original LayerNorm implementation as a component in modalities: https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html

The layer norm was originally instantiated individually for every attention block internally. https://github.com/Modalities/modalities/blob/dd0db07bbe631e9dc30f35912076d26603f4a6b7/src/modalities/models/gpt2/gpt2_model.py#L193

For every new layer norm type we would have had to add an if-clause to check which layer norm we would want to instantiate. As a workaround, we now pass in the layer norm object from outside to the GPT2 model and copy it in every attention block. Note that we override the copy function in the layer norm implementations. 

For the future, it would make sense to have the possibility to instantiate Lists of components. For instance, a GPTModel would have a dependency for a list of attention block. We would specify a single attention block and instantiate the block n times (see `num_instances` in the YAML below). Each attention block would have a dependency for a layer norm and would not have to be copied internally anymore. 

This is an example: 

```yaml
model:
  component_key: model
  variant_key: gpt2
  config:
    [...]
    attention_blocks:
        component_key: attention_block
        variant_key: gpt2_attention_block
        num_instances: 12 
        config:    
          n_embd: 768
          dropout: 0.0
          scaling_factor: 3
         [...]
```


config_files/config_example_mem_map_dataset.yaml
```diff
@@ -31,7 +31,7 @@ train_dataset:
   component_key: dataset
   variant_key: packed_mem_map_dataset_megatron
   config: 
-    raw_data_path: /raid/s3/opengptx/max_lue/LLMgym/data/redpyjama_v2_default_DE_num_docs_16777216.pbin
+    raw_data_path: /raid/s3/opengptx/max_lue/modalities/data/sample_datasets/redpajama_v2/mem_map/redpajama_v2_gpt2_tokenized_num_samples_1050391.pbin
```
flxst, line 34: We should probably use relative paths here (and in other configs, too).

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+class LayerNormIF(nn.Module):
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        raise NotImplementedError
+
+    def __copy__(self):
+        raise NotImplementedError
+
+
+class RMSLayerNorm(LayerNormIF):
+    def __init__(self, ndim: int, epsilon: float = 1e-6):
```
flxst, line 20: Should we not implement an (optional) bias for `RMSLayerNorm`, just like we do for `ZLayerNorm`? The original RMSNorm paper uses a bias by default.

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Initialize the RMSNorm normalization layer.
+        Original paper: https://arxiv.org/pdf/1910.07467.pdf
+        Source code adopted from https://github.com/facebookresearch/llama/blob/a0a4da8b497c566403941ceec47c2512ecf9dd20/llama/model.py#L34C1-L77C36
+
+        Args:
+            ndim (int): The dimension of the input tensor.
+            eps (float, optional): A small value added to the denominator for numerical stability. Default is 1e-6.
```
flxst, line 28: Epsilon is `1e-6` in the LLaMa implementation by default. However, it seems that they actually used `1e-5` themselves, see [here](https://github.com/facebookresearch/llama/issues/771#issuecomment-1761430212). `1e-5` is also the default value in PyTorch for [LayerNorm](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html) and used elsewhere for RMSNorm (e.g. [here](https://huggingface.co/mosaicml/mpt-7b/blob/039e37745f00858f0e01e988383a8c4393b1a4f5/norm.py#L28)), so it seems like a standard value that we should perhaps also use instead of `1e-6`?

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__()
+        self.epsilon = epsilon
+        self.weight = nn.Parameter(torch.ones(ndim))
+
+    def _norm(self, x: torch.Tensor) -> torch.Tensor:
+        return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.epsilon)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        output = self._norm(x.float()).type_as(x)
+        return output * self.weight
```
flxst, line 44: This RMSNorm implementation taken from LLaMa seems to use a plain Euclidean norm, forcing the output vectors to be on a unit sphere. However, in the original RMSNorm paper, there is an additional factor `sqrt(n)` included in the normalisation, where `n` is the dimension of the input/output vectors. This forces the output vectors to be on a `sqrt(n)`-scaled unit sphere instead. The authors claim that this is important. However, I assume that the reason to drop this additional factor is that it can be absorbed in the weights, if `n` is constant?

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def __copy__(self):
+        copied_instance = RMSLayerNorm(self.weight.shape[0], self.epsilon)
+        return copied_instance
+
+
+class ZLayerNorm(LayerNormIF):
```
flxst, line 51: Why is it called `ZLayerNorm`, i.e. what does the `Z` stand for? Is this only to differentiate it from the more generic `LayerNormIF` class?

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+    """LayerNorm but with an optional bias. PyTorch doesn't support simply bias=False"""
```
flxst, line 52: A custom LayerNorm class to support `bias=False` seems to be a relict from [nanoGPT](https://github.com/karpathy/nanoGPT/blob/325be85d9be8c81b436728a420e85796c57dba7e/model.py#L19). It seems that PyTorch supports this natively (see [LayerNorm](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html)) since version 2.1 (see [here](https://github.com/pytorch/pytorch/pull/101683)). Can’t we just use PyTorch’s `LayerNorm` instead of our own `ZLayerNorm` class?

src/modalities/models/components/layer_norms.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ZLayerNormConfig(BaseModel):
+    ndim: Annotated[int, Field(strict=True, ge=1)]
+    bias: Annotated[bool, Field(default=True)]
+    epsilon: Annotated[float, Field(gt=0, default=1e-6)]
+
+
+class RMSLayerNormConfig(BaseModel):
+    ndim: Annotated[int, Field(strict=True, ge=1)]
+    epsilon: Annotated[float, Field(gt=0, default=1e-6)]
```
flxst, line 92: Is there a reason we do not use `strict=True` here (and above in `ZLayerNormConfig`)?

tests/models/components/test_layer_norms.py
```diff
@@ -0,0 +1,47 @@
...
+def test_z_layer_norm_forward(z_layer_norm):
+    x = torch.Tensor([0.1, 0.2, 0.3])
+    output = z_layer_norm(x)
+    ndim = z_layer_norm.weight.shape[0]
+    ref_tensor = F.layer_norm(
+        input=x,
+        normalized_shape=z_layer_norm.weight.shape,
+        weight=nn.Parameter(torch.ones(ndim)),
+        bias=nn.Parameter(torch.zeros(ndim)) if z_layer_norm.bias_tensor is not None else None,
+        eps=z_layer_norm.epsilon,
```
flxst, line 43: I don’t think this test is great because it uses `F.layer_norm()` exactly like `ZLayerNorm` does internally. It would probably be better to do the math by hand and double-check the results this way, similar to what is done in `test_rms_layer_norm_forward()`.

flxst (commented):
Looks good to me! I added a few comments and questions.

src/modalities/models/components/layer_norms.py
BlueCrescent: I think the sqrt(n) factor is part of the mean() call in line 40.

src/modalities/models/components/layer_norms.py
le1nux: The layer norm that is implemented in Pytorch basically calculates the z scores for each vector component (with two additional, learnable affine transformation parameters):
https://en.wikipedia.org/wiki/Standard_score

I found that the naming of layer norm is too generic as RMSNorm is also a layer norm. What naming would you suggest?

src/modalities/models/components/layer_norms.py
le1nux: Good point! I also checked the original RMSNorm implementation and they also had it (see: https://github.com/bzhangGo/rmsnorm/blob/2e726f1a3f106bb719056422f4f9b6aca03d3ce6/rmsnorm_torch.py#L32)

src/modalities/models/components/layer_norms.py
le1nux: Michael and Mehdi suggested to also use the original Layer Norm name. Given the overrulement it's LayerNorm again :-) 
Alsok we don't use a custom LayerNorm wrapper anymore. I found a way to simplify that part so that we don't have to override `__copy__()`.

src/modalities/models/components/layer_norms.py
le1nux: fixed.

mali-git (approved):

fromm-m (approved):

PR #77, opened 2024-03-15
feat: integrate LR schedulers

mali-git:


src/modalities/config/config.py
```diff
@@ -132,9 +134,62 @@ class CheckpointingConfig(BaseModel):
...
+    total_steps: int
+    epochs: Optional[int] = None
+    steps_per_epoch: Optional[int] = None
+    pct_start: float
+    anneal_strategy: str
+    cycle_momentum: bool = True
+    base_momentum: float | List = 0.85
+    max_momentum: float | List = 0.95
+    div_factor: float
+    final_div_factor: float
```
le1nux, line 173: we could also be more restrictive here:

`Annotated[int, Field(strict=True, gt=0)]`

same for the  other int and float types.

le1nux (approved):
Looks good to me! Left some two tiny comments.

src/modalities/config/config.py
mali-git: Done in 8a4702d425cf8f8a4e3e351726c160d0317acf5c.

PR #99, opened 2024-04-04
Independent checkpointing and evaluation period

le1nux:
Allows to define the logging intervals in the config for 
* training loss and troughput
* evaluation 
* checkpointing


src/modalities/trainer.py
mali-git, line 1:
```
def _get_local_sample_id(batch_id: int, batch_size: int) -> int:
        return (batch_id + 1) * batch_size - 1
```

Why do we reduce the batch_size by 1?

src/modalities/util.py
```diff
@@ -21,12 +21,10 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
 
 
 def get_callback_interval_in_batches_per_rank(
```
le1nux, line 23: unused method

mali-git (approved):

PR #100, opened 2024-04-05
Grad norm logging

le1nux:
implemented the logging of the gardient norm *before* the application of gradient clipping


fromm-m (approved):
can be merged

PR #114, opened 2024-04-26
feat: added shell scripts for chunking files into parts and combining…

le1nux:
… the parts to the original file.


mali-git:
My suggestion is to move this to a different repository.

PR #74, opened 2024-03-12
feat: group-query-attention implementation

flxst:
Re-opened version of https://github.com/Modalities/modalities/pull/41.

Potential solution for handling the combination of GQA and FlashAttention: https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

PR #115, opened 2024-04-27
fix: fixed  tensor reshape operations

le1nux:
we experienced loss convergence to almost 0 while inference performance was rather low. 
We figured, there was an information leakage bug due to view and reshape operations that lead to parts of embeddings of different tokens being mixed.

### DIFF
--- a/config_files/training/config_lorem_ipsum.yaml
+++ b/config_files/training/config_lorem_ipsum.yaml
@@ -184,7 +184,7 @@ model:
     vocab_size: 50304 # GPT-2 vocab_size of 50257, padded up to nearest multiple of 64 for efficiency
     n_layer: 2
     n_head_q: 8
-    n_head_kv: 8
+    n_head_kv: 4
     ffn_hidden: 128
     n_embd: 128
     dropout: 0.0
--- a/src/modalities/models/gpt2/gpt2_model.py
+++ b/src/modalities/models/gpt2/gpt2_model.py
@@ -6,8 +6,8 @@ from typing import Annotated, Dict, List, Tuple
 
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
 import xformers.ops as xops
+from flash_attn import flash_attn_func
 from pydantic import BaseModel, Field, model_validator, validator
 
 from modalities.config.config import PydanticPytorchModuleType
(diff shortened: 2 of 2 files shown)


flxst (commented):
Oops, good catch! But why was `flash_attn` replaced by `scaled_dot_product_attention`?

le1nux:
According to the documentation, this is an implementation that is equivalent to flash attention2 but directly shipped with PyTorch as a beta feature. We even saw a 10% throughput increase with the new PyTorch implementation on the DGX boxes.

https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

flxst:
> According to the documentation, this is an implementation that is equivalent to flash attention2 but directly shipped with PyTorch as a beta feature. We even saw a 10% throughput increase with the new PyTorch implementation on the DGX boxes.
> 
> https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html
> 
> cc @fromm-m correct me if I'm mistaken here... :-)

Ok, but note as part of #74, we did the opposite change and went from `scaled_dot_product_attention` to `flash_attn`, see https://github.com/Modalities/modalities/pull/74/commits/7d27b59d9e71ecdba40ea8a72521ad38d361aea7. I wonder what happened to the GQA part, in particular the previously used `repeat_kv` function which took care of the mapping between query heads and key / value heads. Isn't that part missing now? The tests in `tests/models/test_causal_self_attention.py` fail, and I think this might be because pytorch's `scaled_dot_product_attention` does not internally implement GQA, unlike `flash_attn` (see [here](https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#how-to-use-flashattention)). Or am I missing something?

fromm-m:
> flash_attn

You are right, flash-attention in PyTorch still does not support GQA, with the newest changes we are now back to flash_attn from Daio for now.

fromm-m:
> > According to the documentation, this is an implementation that is equivalent to flash attention2 but directly shipped with PyTorch as a beta feature. We even saw a 10% throughput increase with the new PyTorch implementation on the DGX boxes.
> > https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html
> > cc @fromm-m correct me if I'm mistaken here... :-)
> 
> Ok, but note that as part of #74, we did the opposite change and went from `scaled_dot_product_attention` to `flash_attn`, see [7d27b59](https://github.com/Modalities/modalities/commit/7d27b59d9e71ecdba40ea8a72521ad38d361aea7). I wonder what happened to the GQA part, in particular the previously used `repeat_kv` function which took care of the mapping between query heads and key / value heads. Isn't that part missing now? The tests in `tests/models/test_causal_self_attention.py` fail, and I think this might be because pytorch's `scaled_dot_product_attention` does not internally implement GQA, unlike `flash_attn` (see [here](https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#how-to-use-flashattention)). Or am I missing something?

You are right, flash-attention in PyTorch still does not support GQA, with the newest changes we are now back to flash_attn from Daio for now.

fromm-m (commented):
can be merged

PR #116, opened 2024-04-29
11 implement moe

AbasKhan:


lllAlexanderlll (commented):
Great work! Well done! :)
For me, the tests run through locally, but github actions seems to have a problem with installing modalities for the  tests.
When running `pre-commit run --all-files` locally, two files will be altered. This should get fixed, please.

A clarification question:
In this initial MoE implementation, we could face inefficient training where only a few lucky experts get trained intensively, as there is no auxiliary loss, expert capacity or similar technique implemented, yet, right?
E.g. as mentioned here https://huggingface.co/blog/moe#load-balancing-tokens-for-moes

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+        self.moe_num_experts = moe_config.moe_num_experts
+        self.moe_top_k = moe_config.moe_top_k
+        self.moe_normalize_expert_weights = moe_config.moe_normalize_expert_weights
+        self.uniform_expert_assignment = moe_config.uniform_expert_assignment
+        self.moe_jitter_eps = moe_config.moe_jitter_eps
+
+        self.layer = nn.Linear(self.hidden_size, self.moe_num_experts, bias=False)
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.LongTensor]:
+        if self.training and self.moe_jitter_eps is not None:
```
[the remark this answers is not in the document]
AbasKhan: I think so too, i really liked the "gap text" analogy/explanation, nice way of visualizing whats going on . Plus the first dimension i think is more being combined with the sequence length (sounds more accurate than getting rid i guess) to create a batch of sequence elements, each of which will be independently routed to the experts.

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+
+        weights = self.layer(x.view(-1, x.shape[-1])).softmax(dim=-1, dtype=torch.float32)
+        top_weights, top_experts = torch.topk(weights, self.moe_top_k, dim=-1)
+
+        if self.moe_normalize_expert_weights:
+            top_weights = top_weights / torch.norm(
+                top_weights, p=self.moe_normalize_expert_weights, dim=-1, keepdim=True
+            )
+
+        if self.uniform_expert_assignment:
```
mrudat-iais, line 43: then the top_k experts are chosen uniformly but the returned weights stay as they are, right?

in which cases would we do that?

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+    def _jitter(self, x: torch.Tensor) -> torch.Tensor:
+        if self.moe_jitter_eps is None:
+            raise RuntimeError("The router does not have moe_jitter_eps set.")
+        low = 1.0 - self.moe_jitter_eps
+        high = 1.0 + self.moe_jitter_eps
+        noise = torch.rand(x.size(), dtype=x.dtype, device=x.device)
+        return low + noise * (high - low)
+
+
+class MoEExpertGLU(nn.Module):
```
mrudat-iais, line 64: GLU = Gated Linear Unit ?
Does that mean each Expert is a GLU layer?
Why is it not just e.g. a FF Layer?

tests/nn/test_moe.py
```diff
@@ -0,0 +1,121 @@
...
+
+def test_moe_expert_produces_expected_shape(
+    model_input: torch.Tensor, batch_size: int, seq_length: int, hidden_size: int, moe_num_experts: int
+):
+    ffn_hidden_size = 128
+    act_fn = nn.ReLU
+    model = MoEExpertGLU(hidden_size, ffn_hidden_size, moe_num_experts, act_fn)
+    expert_idx = 2
+    output = model.forward(model_input, expert_idx)
+    assert output.shape == (batch_size, seq_length, hidden_size)
```
mrudat-iais, line 33: The output has the correct shape. Is it intended that the output is filled with nan values? This could be because of the random test configs maybe?

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+        self, hidden_size: int, ffn_hidden_size: int, moe_num_experts: int, act_fn: Callable[[], nn.Module] = nn.GELU
+    ):
+        super().__init__()
+        self.moe_num_experts = moe_num_experts
+        self.mlp = MoEExpertGLU(
+            hidden_size=hidden_size, ffn_hidden_size=ffn_hidden_size, moe_num_experts=moe_num_experts, act_fn=act_fn
+        )
+
+    def forward(self, x: torch.Tensor, top_weights: torch.Tensor, top_experts: torch.LongTensor) -> torch.Tensor:
+        bsz, q_len, hidden_size = x.shape
```
mrudat-iais, line 102:
```suggestion
        batch_size, sequence_length, hidden_size = x.shape
```

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(
+        self, hidden_size: int, ffn_hidden_size: int, moe_num_experts: int, act_fn: Callable[[], nn.Module] = nn.GELU
+    ):
+        super().__init__()
+        self.moe_num_experts = moe_num_experts
+        self.mlp = MoEExpertGLU(
+            hidden_size=hidden_size, ffn_hidden_size=ffn_hidden_size, moe_num_experts=moe_num_experts, act_fn=act_fn
+        )
+
+    def forward(self, x: torch.Tensor, top_weights: torch.Tensor, top_experts: torch.LongTensor) -> torch.Tensor:
```
mrudat-iais, line 101: Would be nice to go through this together. We were just debugging but did not get the fill picture of what happens here.

mrudat-iais (changes requested):
The code was very readable, especially with the tests you added. We left comments regarding implementation changes. 
Besides that, the MoEExperts.forward() is still unclear to us. Would be nice to go through the code there together ;-)

src/modalities/nn/moe.py
AbasKhan: Danke danke. As for you question, Exactly, if the uniform_expert_assignment is set to True, indeed the experts are chosen uniformly,  this approach in my understanding is useful for load balancing ( there are no favorite experts, favorite experts can both hider the training of other experts and also create a bottleneck on themselves ).  Setting this flag to True though does sort of ignore the router features, so probably isnt best practice. I wonder if combining the two settings makes sense, initially have expert assignment be uniform and only later when the experts have had chance to learn this flag is switched to False.

src/modalities/nn/moe.py
AbasKhan: Probably the naming choice was not the best on our part, GLU is for [Gated Linear Unite] (https://paperswithcode.com/method/glu).  the operation x1 = x1 * x2 is the gating mechanism of the GLU, with x2 being the gate, which control how much information/features form x1 are retained. I am speculating here, but i think this makes the training more stable , and nicely ties into the concept that each expert "specializes" on different features.

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -380,34 +471,17 @@ def __init__(
...
             raise TypeError(f"{poe_type} not supported")
 
-        if poe_type is not PositionTypes.NOPE and RotaryTransform in [
-            config.type_hint.value for config in attention_config.qkv_transforms
-        ]:
-            raise ValueError('It is expected to use "RotaryTransform" together with "NOPE".')
+        # if poe_type is not PositionTypes.NOPE and RotaryTransform in [
+        #     config.type_hint.value for config in attention_config.qkv_transforms
+        # ]:
+        #     raise ValueError('It is expected to use "RotaryTransform" together with "NOPE".')
```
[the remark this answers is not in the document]
AbasKhan: We had commented out this part during the development, I was re-added the check in the code.

src/modalities/nn/moe.py
mrudat-iais: Okay, thanks for clarifying. I guess the initialization in random / uniformly anyways. In the configs the value is set to True. Is that intended?

src/modalities/nn/moe.py
```diff
@@ -0,0 +1,141 @@
...
+        return x1
+
+
+class MoEExperts(nn.Module):
+    def __init__(
+        self, hidden_size: int, ffn_hidden_size: int, moe_num_experts: int, act_fn: Callable[[], nn.Module] = nn.GELU
+    ):
+        super().__init__()
+        self.moe_num_experts = moe_num_experts
+        self.mlp = MoEExpertGLU(
```
rrutmann, line 97: Is mlp a good variable name here? It is also used in gpt2_model.py for MoEFFN

src/modalities/nn/moe.py
AbasKhan: Thanks for pointing this out, For initial phase we wanted to setup the basic architecture of experts so this was kept as True, We are testing/working on adding the dynamic routing part.

src/modalities/nn/moe.py
AbasKhan: Totally agreed that name is probably not the best. We just wanted to be consistent with the names already used, as you can see here https://github.com/Modalities/modalities/blob/2b562b1e08871c67b7cb89ac575aae276e9e9b78/src/modalities/models/gpt2/gpt2_model.py#L325 .

mrudat-iais (approved):
nice work!

Issue #60, opened 2024-02-25
CLM CrossEntropyLoss seems to depend on 


le1nux:
I ran the same config with different two different batch sizes (2 and 16). 
For a batch size of 2, the initial loss is significantly higher than for the higher batch size. There seems to be a dependency between the batch size and how we normalise the CLM CrossEntropyLoss. 

![image](https://github.com/Modalities/modalities/assets/2804731/f8a78f09-c869-4fd9-8690-eccbb84c2723)

PR #118, opened 2024-04-29
refactor: fixed loss logging during gradient accumulation and gradient clipping

le1nux:
Previously with gradient accumulation, the loss was divided by the number of gradient accumulation steps. This normalised loss was then logged. From a logging perspective,  each train step is equivalent to one gradient accumulation step and therefore the normalisation was not needed. 

Additionally, I refactored the gradient clipping in two ways. Firstly, we were not using the FSDP gradient clipping. This is now integrated but needs to be made configurable. This I will do in another PR. 

Secondly, the gradient clipping is only calculated and logged after # gradient accumulation steps. Since garden clipping already syncs the gradient norm across all the ranks, we don't have to apply the reduce operation on them anymore.

### DIFF
--- a/src/modalities/trainer.py
+++ b/src/modalities/trainer.py
@@ -4,6 +4,7 @@ from typing import Callable, Tuple
 import torch
 import torch.distributed as dist
 import torch.nn as nn
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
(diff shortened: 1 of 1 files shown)


fromm-m (commented):
can be merged

le1nux:
fixes #60

PR #135, opened 2024-05-22
Include torch compile

flxst:
Addresses #133.


le1nux:
Already added as part of #312

PR #136, opened 2024-05-23
Fix Pre-LN for GPT2 Model

flxst:
This PR fixes the pre-LN by reverting the changes (accidentally?) introduced [here]( https://github.com/Modalities/modalities/commit/8dd74a4f98968be85c1600200cb6fd0aa5700553#diff-7b510b8da861fb91767d9aefe8db60e8ee5ae624aec986b7dc36da5f34ed48adL207)

References:
- Fig.1 & Tab. 1 in the paper [On Layer Normalization in the Transformer Architecture](https://arxiv.org/abs/2002.04745)
- [nanoGPT](https://github.com/karpathy/nanoGPT/blob/master/model.py#L103)


le1nux (approved):
LGTM. Good find!

PR #138, opened 2024-05-27
Implementation of multiple attention mechanisms

flxst:
This PR implements `manual attention` and `pytorch flash attention`, in addition to the previously implemented `dao flash attention`. Group Query Attention is supported.

### DIFF
--- a/tests/models/test_causal_self_attention.py
+++ b/tests/models/test_causal_self_attention.py
@@ -1,12 +1,20 @@
+"""
+Note: test_attention_types_approximate_equality can print the output of different attention implementations. 
+      To do so, turn on verbose and run 'python tests/models/test_causal_self_attention.py -s'
+"""
+from copy import deepcopy
+
 import pytest
 import torch
 
 from modalities.models.gpt2.gpt2_model import AttentionConfig, CausalSelfAttention
 
+torch.manual_seed(0)
+
 
 def _get_random_input_seq(embedding_shape):
     flash_attn_supported_dtype = torch.bfloat16
-    return torch.rand(size=embedding_shape, dtype=flash_attn_supported_dtype)
+    return torch.rand(size=embedding_shape, dtype=flash_attn_supported_dtype).cuda()
 
 
 def _get_random_attention_layer(n_head_q, n_head_kv, n_embd, block_size, attention_impl, attention_config):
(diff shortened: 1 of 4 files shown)


tests/models/test_causal_self_attention.py
```diff
@@ -26,27 +34,80 @@ def _get_random_attention_layer(n_head_q, n_head_kv, n_embd, block_size, attenti
...
+        (32, 4, 768),  # GQA
+    ],
+)
+def test_repeat_kv_heads(n_head_q, n_head_kv, n_embd):
+    batch_size = 2
+    block_size = 10
+    head_dim = n_embd // n_head_q
+    AttentionConfig(qkv_transforms=[])
+
+    q = torch.rand(batch_size, n_head_q, block_size - 1, head_dim, dtype=torch.bfloat16).cuda()
```
le1nux, line 53: As an idea: instead of torch.rand we could also do `torch.arange(0,batch_size*n_head_q*(block_size-1)*head_dim).reshape(batch_size, n_head_q, block_size - 1, head_dim)`. In this case, we can check for equality instead of approx. equality.

tests/models/test_causal_self_attention.py
```diff
@@ -60,14 +121,102 @@ def test_forward_pass_success(n_head_q, n_head_kv, n_embd, successful):
...
+    output_tensor[attention_impl_1] = attention_layer[attention_impl_1](embedded_input_seq)
+    output_tensor[attention_impl_2] = attention_layer[attention_impl_2](embedded_input_seq)
+    if verbose:
+        print(f"{attention_impl_1} vs. {attention_impl_2}: \n{output_tensor}")
+    torch.testing.assert_close(
+        output_tensor[attention_impl_1],
+        output_tensor[attention_impl_2],
+        atol=2.5e-3,  # default for bfloat16: 1e-5
+        rtol=0.016,  # default for bfloat16: 0.016
+    )
```
le1nux, line 222: We could add another test in which we test the output of the  manual attention implementation against the precomputed output (pen and paper) given a very short input sequence. In this case, we can be entirely sure that the implementation is correct.

le1nux (approved):

le1nux (commented):
Left some minor comments. But I think we can merge it already! Really nice how we can switch now between the different attention implementations and how high test coverage. Nice work! :)

mali-git (approved):

PR #139, opened 2024-05-30
Implementation of optimizer parameter groups with and without weight decay

flxst:
This PR allows to exclude certain groups of parameters (e.g. embeddings, layer norms) from weight decay. The parameter groups are model-dependent and defined as part of the respective model class. Exclusion can be triggered in the training config file.


le1nux (approved):
LGTM :) committed some minor improvements. 
I think we can merge it after @mali-git  or @fromm-m have also reviewed the PR. 

For later, we should add a test that checks the correctness of the parameter groups when using FSDP with fullshard mode (in addition to the noshard test). What do you think @flxst?

fromm-m (approved):
I could not spot a bug! Can be merged.

mali-git (commented):
Looks good!

PR #152, opened 2024-06-11
Tokenizer remove max length flag

le1nux:
By default, we do not specify the tokenizer's `max_length` anymore and set `truncation` and `padding` to false now.


src/modalities/tokenization/tokenizer_wrapper.py
```diff
@@ -22,9 +22,15 @@ def get_token_id(self, token: str) -> int:
 
 class PreTrainedHFTokenizer(TokenizerWrapper):
     def __init__(
-        self, pretrained_model_name_or_path: str, max_length: int, truncation: bool = True, padding: str = "max_length"
+        self,
+        pretrained_model_name_or_path: str,
+        truncation: bool = False,
+        padding: bool = False,
+        max_length: int = None,
```
flxst, line 29:
```suggestion
        max_length: Optional[int] = None,
```
and import `Optional` in line 2

flxst (approved):
The changes look good to me! I only found a single typing issue. 

However, I think that the unit tests (`test_hf_tokenize`) do not suffice. Maybe it would be a good idea to explicitly check the tokenization of a multi-document example, and vary parameters like `truncation`, `padding` and `max_length`?

le1nux:
I added more test cases concerning the relevant combination of  `max_length`, `padding`,  `truncation` for the single document case. 
The multi-document case it currently not supported and not needed so far, see [def tokenize(self, text: str) -> List[int]:](https://github.com/Modalities/modalities/blob/3ba062b763ce9d0fef04760863964afb88672104/src/modalities/tokenization/tokenizer_wrapper.py#L57)

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
+                pretrained_model_name_or_path="data/tokenizer/hf_gpt2",
+                truncation=True,
+                padding=True,
+                max_length=None,
+                special_tokens={"pad_token": "[PAD]"},
+            ),
+            15,
+            0,
+        ),
+        # If truncation=False and len(text)<model max length, we want a sequence to be truncated to the max_length, irrespective of the padding flag.
```
flxst, line 175:
```suggestion
        # If truncation=False and len(text)<model max length, we want a sequence to be unmodified, irrespective of the padding flag.
```

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
+
+    tokenizer_config_dict = tokenizer_config.model_dump()
+    tokenizer = PreTrainedHFTokenizer(**tokenizer_config_dict)
+
+    token_ids = tokenizer.tokenize(text)
+
+    # make sure that the overall token sequence length is correct
+    assert len(token_ids) == expected_length
+
+    # check number of non-padding tokens
```
flxst, line 308:
```suggestion
    # check number of non-padding tokens (token_id = 43488 corresponds to "AAAAAAAA")
```

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
+            1024,
+            4,
+        ),
+    ],
+)
+def test_hf_tokenize(
+    text: str,
+    tokenizer_config: PreTrainedHFTokenizerConfig,
+    expected_length: int,
+    expected_num_paddding_tokens: int,
```
flxst, line 295:
```suggestion
    expected_num_padding_tokens: int,
```
Typo, occurrence 1 of 3

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
+    tokenizer = PreTrainedHFTokenizer(**tokenizer_config_dict)
+
+    token_ids = tokenizer.tokenize(text)
+
+    # make sure that the overall token sequence length is correct
+    assert len(token_ids) == expected_length
+
+    # check number of non-padding tokens
+    assert sum(np.array(token_ids) == 43488) == (
+        expected_length - expected_num_paddding_tokens
```
flxst, line 310:
```suggestion
        expected_length - expected_num_padding_tokens
```
Typo, occurrence 2 of 3

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
     )
-    _assert_tokenization(tokenizer)
+
+    # check number of padding tokens
+    assert sum(np.array(token_ids) == 50257) == expected_num_paddding_tokens
```
flxst, line 314:
```suggestion
    assert sum(np.array(token_ids) == 50257) == expected_num_padding_tokens
```
Typo, occurrence 3 of 3

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
-    tokenizer = PreTrainedHFTokenizer(
-        pretrained_model_name_or_path=tokenizer_model_file, max_length=20, truncation=False, padding=False
+@pytest.mark.parametrize(
+    "text,tokenizer_config,expected_length,expected_num_paddding_tokens",
+    [
+        # Test cases 1: Sequence is shorter than max_length, i.e., len(text) < max_length
+        # If padding="max_length", we want a sequence to be padded to the max_length, irrespective of the truncation flag
+        # and only if max_length is specified.
+        # If max_length is not specified, we pad to the max model input length (i.e., 1024 for the gpt2 model).
+        # NOTE: "AAAAAAAA" is a single token for the gpt2 tokenizer, there is "A" sequence longer than that in the vocabulary
```
flxst, line 25: I do not understand the second part of the comment in line 25. Also, if this is a general comment not specific to test cases 1, it should perhaps be moved up.

tests/test_tokenization.py
```diff
@@ -9,12 +15,303 @@ def _assert_tokenization(tokenizer: TokenizerWrapper):
...
+            PreTrainedHFTokenizerConfig(
+                pretrained_model_name_or_path="data/tokenizer/hf_gpt2",
+                truncation=False,
+                padding="max_length",
+                max_length=10,
+                special_tokens={"pad_token": "[PAD]"},
+            ),
+            10,
+            4,
+        ),
```
flxst, line 37: The `PreTrainedHFTokenizerConfig` is used with the same arguments `pretrained_model_name_or_path` and `special tokens` for all test cases. It would perhaps be simpler and more readable if only `truncation`, `padding` and `max_length` were used as input arguments, with `PreTrainedHFTokenizerConfig` being instantiated in the test function below.

src/modalities/tokenization/tokenizer_wrapper.py
```diff
@@ -22,9 +22,26 @@ def get_token_id(self, token: str) -> int:
 
 class PreTrainedHFTokenizer(TokenizerWrapper):
     def __init__(
-        self, pretrained_model_name_or_path: str, max_length: int, truncation: bool = True, padding: str = "max_length"
+        self,
+        pretrained_model_name_or_path: str,
+        truncation: bool = False,
+        padding: bool = False,
```
flxst, line 28:
```suggestion
        padding: bool | str = False,
```

flxst (approved):
Nice work! Looks much better than before. I left a few minor comments and suggestions for improvement.

tests/test_tokenization.py
le1nux:
```suggestion
        # NOTE: "AAAAAAAA" is a single token for the gpt2 tokenizer, there is no "A" sequence longer than that in the vocabulary.
```

le1nux: was missing a "no". good find!

PR #157, opened 2024-06-17
Fix CPU Tests

flxst:
This PR fixes the CPU tests with github actions


pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 requires-python = ">=3.8,<3.12"
 description = "Modalities, a python framework for distributed and reproducible foundation model training."
 dependencies = [
+    "numpy<2.0",
```
le1nux, line 7: why do we need the nump version to be `<2.0` ?

le1nux (approved):
Apart from my comment regarding numpy version, all looks good. 
Also, all CPU tests run through again! 🎉

pyproject.toml
flxst: Numpy 2.0 introduces breaking changes and is currently not compatible with the latest version of at least one of our dependencies (`accelerate`). It leads to the following error: 

```AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.. Did you mean: 'float16'?```

le1nux (approved):

PR #158, opened 2024-06-19
Fix/sequence length power of 2

le1nux:
# What does this PR do?
Previously, the `block_size` in the dataset would be set to a power of two, resulting in the sequence length being `block_size -1`, which is not best practice and can impact the model training e.g., throughput-wise. 

As a fix, we now specify the `sequence_length` in the config instead of the `block_size`. During Dataset instantiation we chose the `block_size` to be `sequence_length+1`.

Previously, we would also chunk the dataset into `block_size` long chunks. Each chunk would then be used for training individually. As a result, the last token of a block would be only used as a target but never as an input. We changed this, such that we reuse the last token of a batch as the first one of the subsequent batch. 

## General changes
* nothing apart from points mentioned above

## Breaking Changes
* replaced `block_size` in `Dataset`, `Model` and `NumberConversion` with `sequence_length`

- [ ] I have merged main into this feature branch


src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -519,7 +518,9 @@ def forward_impl(self, inputs: Dict[str, torch.Tensor]) -> Dict[str, torch.Tenso
         input_ids = inputs[self.sample_key]
         device = input_ids.device
         b, t = input_ids.size()  # batch size, sequence length
-        assert t <= self.block_size, f"Cannot forward sequence of length {t}, block size is only {self.block_size}"
+        assert (
+            t <= self.sequence_length
+        ), f"Cannot forward sequence of length {t}, block size is only {self.sequence_length}"
```
flxst, line 523:
```suggestion
        ), f"Cannot forward sequence of length {t}, the model's maximum input sequence length is only {self.sequence_length}"
```

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -519,7 +518,9 @@ def forward_impl(self, inputs: Dict[str, torch.Tensor]) -> Dict[str, torch.Tenso
         input_ids = inputs[self.sample_key]
         device = input_ids.device
         b, t = input_ids.size()  # batch size, sequence length
```
flxst, line 520:
```suggestion
        _, t = input_ids.size()  # batch size, sequence length
```
batch size `b` is not used

tests/checkpointing/test_fsdp_to_disc_checkpointing.py
```diff
@@ -111,9 +111,9 @@ def _clone_parameters(fsdp_wrapped_model):
     def _generate_batch(gpt2_model_config: Dict):
         # prepare input and targets
         data = torch.randint(
-            0,
-            gpt2_model_config["model"]["config"]["vocab_size"],
-            (8, gpt2_model_config["model"]["config"]["block_size"] + 1),
+            0,  # lowest token_id
+            gpt2_model_config["model"]["config"]["vocab_size"],  # highest token_id, i.e, vocab_size
```
flxst, line 115:
```suggestion
            gpt2_model_config["model"]["config"]["vocab_size"],  # highest token_id + 1, i.e. vocab_size
```

flxst (approved):
Great work! Added some minor comments.

mali-git (approved):

Issue #32, opened 2024-01-28
Activation Checkpointing has hardcoded GPT2Block


le1nux:
We should be able to set the modules that are subject to activation checkpointing via the config

Currently, it is hardcoded:
```python
def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module):
    return isinstance(submodule, GPT2Block)
```

see: https://github.com/Modalities/modalities/blob/main/src/modalities/activation_checkpointing.py#L15

PR #159, opened 2024-06-24
Make Activation Checkpointing Configurable

mali-git:
This PR addresses the issue described in [Issue #32](https://github.com/Modalities/modalities/issues/32) by removing the hard-coded dependency for applying activation checkpointing specifically to `GPT2Block`. Instead, activation checkpointing is now made configurable through the configuration file, allowing greater flexibility and adaptability for different models.

**General changes:**
* Enhancement: Allow the definition of modules to be checkpointed within the configuration file

**Breaking changes:** 
* Enhacement: make modules to be checkpointed configurable.
This requires replacing the boolean entry `do_apply_activation_checkpointing` in the configuration file with the list `activation_checkpointing_modules` (see 4a2a049e893acdd49c2dd7801b74a20af0db1e38), which specifies all the submodules to which activation checkpointing should be applied.

- [x] I have merged main into this feature branch

### DIFF
--- a/src/modalities/activation_checkpointing.py
+++ b/src/modalities/activation_checkpointing.py
@@ -1,4 +1,6 @@
 from functools import partial
+from typing import List
+from accelerate import FullyShardedDataParallelPlugin
 
 import torch
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
@@ -8,14 +10,16 @@ from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
 )
 from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
 
-from modalities.models.gpt2.gpt2_model import GPT2Block
+
+import torch
+from typing import List
+
+def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module, activation_checkpointing_modules: List[type]) -> bool:
+    return isinstance(submodule, tuple(activation_checkpointing_modules))
 
 
-def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module) -> bool:
-    return isinstance(submodule, GPT2Block)
-
-
-def apply_activation_checkpointing_inplace(model: torch.nn.Module):
+def apply_activation_checkpointing_inplace(model: torch.nn.Module, activation_checkpointing_modules: List[str]):
+    activation_checkpointing_modules = [FullyShardedDataParallelPlugin.get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
     assert isinstance(model, FSDP), "activation checkpointing can only be applied to FSDP wrapped models!"
     non_reentrant_wrapper = partial(checkpoint_wrapper, checkpoint_impl=CheckpointImpl.NO_REENTRANT, debug=False)
 
--- a/src/modalities/activation_checkpointing.py
+++ b/src/modalities/activation_checkpointing.py
@@ -21,7 +21,8 @@ def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module, acti
 
 def apply_activation_checkpointing_inplace(model: torch.nn.Module, activation_checkpointing_modules: List[str]):
     activation_checkpointing_modules = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
-    assert isinstance(model, FSDP), "activation checkpointing can only be applied to FSDP wrapped models!"
+    if not isinstance(model, FSDP)
+        raise ValueError("activation checkpointing can only be applied to FSDP wrapped models!")
     non_reentrant_wrapper = partial(checkpoint_wrapper, checkpoint_impl=CheckpointImpl.NO_REENTRANT, debug=False)
 
     apply_activation_checkpointing(
--- a/src/modalities/activation_checkpointing.py
+++ b/src/modalities/activation_checkpointing.py
@@ -1,6 +1,5 @@
 from functools import partial
 from typing import List
-from accelerate import FullyShardedDataParallelPlugin
 
 import torch
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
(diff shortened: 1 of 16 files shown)


src/modalities/activation_checkpointing.py
```diff
@@ -8,14 +9,18 @@
...
-def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module) -> bool:
-    return isinstance(submodule, GPT2Block)
+def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module, activation_checkpointing_modules: List[type]) -> bool:
+    return isinstance(submodule, tuple(activation_checkpointing_modules))
 
 
-def apply_activation_checkpointing_inplace(model: torch.nn.Module):
+def apply_activation_checkpointing_inplace(model: torch.nn.Module, activation_checkpointing_modules: List[str]):
+    activation_checkpointing_modules = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
     assert isinstance(model, FSDP), "activation checkpointing can only be applied to FSDP wrapped models!"
```
le1nux, line 24:
```suggestion
    if not isinstance(model, FSDP)
        raise ValueError("activation checkpointing can only be applied to FSDP wrapped models!")
```

asserts normally should not be used in production code, as these checks can be deactivated in optimised code. 

see here: https://stackoverflow.com/questions/56990925/is-this-assert-for-development-not-for-production

src/modalities/activation_checkpointing.py
```diff
@@ -8,14 +9,18 @@
...
 
-def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module) -> bool:
-    return isinstance(submodule, GPT2Block)
+def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module, activation_checkpointing_modules: List[type]) -> bool:
+    return isinstance(submodule, tuple(activation_checkpointing_modules))
 
 
-def apply_activation_checkpointing_inplace(model: torch.nn.Module):
+def apply_activation_checkpointing_inplace(model: torch.nn.Module, activation_checkpointing_modules: List[str]):
+    activation_checkpointing_modules = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
```
le1nux, line 23:
```suggestion
    activation_checkpointing_module_types = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
```

le1nux: otherwise, this might be confused with actual instances. Here, we are just passing model types around. 

(also other occurences of `activation_checkpointing_modules` would need renaming)

src/modalities/config/instantiation_models.py
```diff
@@ -32,7 +32,7 @@ class Training(BaseModel):
             global_training_log_interval_in_steps: Annotated[int, Field(strict=True, ge=1)]
             global_checkpointing_interval_in_steps: Annotated[int, Field(strict=True, ge=1)]
             global_evaluation_interval_in_steps: Annotated[int, Field(strict=True, ge=1)]
-            do_apply_activation_checkpointing: bool
+            activation_checkpointing_modules: Optional[List[str]]
```
le1nux, line 35: default value missing, as we discussed yesterday

le1nux (changes requested):
Looks good to me. Just realised in the middle of the review that the changes from yesterday's discussion are missing. Nevertheless I continued and only found some minor things that we could potentially improve.

Let me know once the changes are in and I will take another look.

src/modalities/config/instantiation_models.py
mali-git: Done in fa1cca2ff1a4303e1b528641b61aa4d634679cac.

src/modalities/activation_checkpointing.py
```diff
@@ -20,10 +20,11 @@ def is_module_to_apply_activation_checkpointing(submodule: torch.nn.Module, acti
...
-    activation_checkpointing_modules = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
-    assert isinstance(model, FSDP), "activation checkpointing can only be applied to FSDP wrapped models!"
+    activation_checkpointing_module_types = [get_module_class_from_name(model, m) for m in activation_checkpointing_modules]
+    if not isinstance(model, FSDP)
+        raise ValueError("activation checkpointing can only be applied to FSDP wrapped models!")
     non_reentrant_wrapper = partial(checkpoint_wrapper, checkpoint_impl=CheckpointImpl.NO_REENTRANT, debug=False)
 
     apply_activation_checkpointing(
-        model, checkpoint_wrapper_fn=non_reentrant_wrapper, check_fn=is_module_to_apply_activation_checkpointing
+        model, checkpoint_wrapper_fn=non_reentrant_wrapper, check_fn=lambda submodule: is_module_to_apply_activation_checkpointing(submodule, activation_checkpointing_modules)
```
le1nux, line 29:
```suggestion
        model, checkpoint_wrapper_fn=non_reentrant_wrapper, check_fn=lambda submodule: is_module_to_apply_activation_checkpointing(submodule, activation_checkpointing_modules_types)
```
shouldn't it be that way?

le1nux, line 29: The `non_reentrant_wrapper` is a partial function. We could do the same to replace the lambda expression.

le1nux (changes requested):
Left two remarks. One seems to be a copy and paste error. 

Looks good otherwise and once addressed we can merge from my end. 

One more review from @flxst or @fromm-m  would be great through.

le1nux:
LGTM :-) Nice work!

le1nux (approved):

fromm-m (approved):
we can still remove accelerate from the pytorch.toml as a dependency but besides that it looks great!

mali-git:
> we can still remove accelerate from the pytorch.toml as a dependency but besides that it looks great!

Done in 8605dbecbe46c068668f283bf4a6352c85ca1c21.

PR #160, opened 2024-06-24
Generic downstream eval

mrudat-iais:
# What does this PR do?
We can now convert a modalities model into a HF model. This model can then be evaluated in eval harness. 
* (**main**) We now have a `convert_pytorch_to_hf_checkpoint` endpoint that takes the `model_name`, `model_config_path` and `output_dir` and creates HF files that can then be loaded with `AutoModel.from_pretrained`
* (**tests**) we have a test that shows the functionality [tests/checkpointing/test_checkpoint_conversion.py](https://github.com/Modalities/modalities/blob/019b8074f79efefa484c580824b4bf9b60ace5da/tests/checkpointing/test_checkpoint_conversion.py)

## General changes
* none besides the above
* PR feedback addressed and implemented

## Breaking Changes
* no breaking changes

- [x] I have merged main into this feature branch


le1nux (commented):
Really good work and awesome that you guys found a generic solution to it! :-) Thanks a lot for digging deep into the huggingface source code!

I left a few remarks here and there that you should address but overall the solution is great!

le1nux (changes requested):
Really good work and awesome that you guys found a generic solution to it! :-) Thanks a lot for digging deep into the huggingface source code!

I left a few remarks here and there that you should address but overall the solution is great!

le1nux (commented):
Really good work and awesome that you guys found a generic solution to it! :-) Thanks a lot for digging deep into the huggingface source code!

I left a few remarks here and there that you should address but overall the solution is great!

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,87 @@
...
+from modalities.models.utils import get_model_from_config
+
+
+class HFAdapterConfig(PretrainedConfig):
+    model_type = "modalities"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if self.config:
+            self.convert_posixpath_to_str(self.config)
```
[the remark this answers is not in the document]
rrutmann: We added an explanation, but decided to omit the example since it's not a complex operation

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        Recursively iterate over the dictionary and convert PosixPath values to strings.
+        """
+        for key, value in d.items():
+            if isinstance(value, PosixPath):
+                d[key] = str(value)
+            elif isinstance(value, dict):
+                self.convert_posixpath_to_str(value)
+            elif isinstance(value, list):
+                d[key] = [str(item) if isinstance(item, PosixPath) else item for item in value]
```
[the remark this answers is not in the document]
rrutmann: These are `PosixPaths`s from our config

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,87 @@
...
+
+class HFAdapterConfig(PretrainedConfig):
+    model_type = "modalities"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if self.config:
+            self.convert_posixpath_to_str(self.config)
+
+    def to_json_string(self, use_diff: bool = True) -> str:
```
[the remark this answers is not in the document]
rrutmann: The override decorator is only supported by python 3.12 or newer, but in our pyproject.toml we actually require python < 3.12

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,87 @@
...
+            elif isinstance(value, list):
+                d[key] = [str(item) if isinstance(item, PosixPath) else item for item in value]
+
+
+class HFAdapter(PreTrainedModel):
+    config_class = HFAdapterConfig
+
+    def __init__(self, config: HFAdapterConfig, model: Optional[nn.Module] = None, *inputs, **kwargs):
+        super().__init__(config)
+        if not model:
```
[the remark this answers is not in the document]
rrutmann: We removed the model-parameter and now always instantiate the model internally

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,87 @@
...
+        output_hidden_states: Optional[bool] = False,
+    ):
+        if output_attentions or output_hidden_states:
+            raise NotImplementedError
+        model_input = {"input_ids": input_ids, "attention_mask": attention_mask}
+        model_forward_output: Dict[str, torch.Tensor] = self.model.forward(model_input)
+        if return_dict:
+            return ModalitiesModelOutput(**model_forward_output)
+        else:
+            return model_forward_output[self.model.prediction_key]
```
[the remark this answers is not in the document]
rrutmann: We think in our case it is necessary that the model has a prediction key. Else how should we know which entry to access in model_forward_output? We add therefore an explicit assertion to ensure that model.prediction_key exists

tests/checkpointing/test_checkpoint_conversion.py
```diff
@@ -0,0 +1,129 @@
...
+    return test_tensor
+
+
+def test_hf_and_pytorch_models_are_the_same_after_init(hf_model, pytorch_model, checkpoint_conversion):
+    assert hf_model.dtype == pytorch_model.lm_head.weight.dtype
+    assert os.listdir(checkpoint_conversion.output_hf_checkpoint_dir)
+
+
+def test_models_before_and_after_conversion_produce_same_output(
+    device,
```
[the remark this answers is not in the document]
rrutmann: Do we really need type hints for test fixtures? Isn't it enough to have the return type for each fixture?

src/modalities/models/huggingface_adapters/hf_adapter.py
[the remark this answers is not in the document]
le1nux: true, I also saw this today. We can keep it for later version of modalities. :-)

src/modalities/models/huggingface_adapters/hf_adapter.py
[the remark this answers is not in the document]
le1nux: The issue is that `NNModel`does not have the attribute `prediction_key` and cannot assume that every inherited model has it. So by "contract" of the class definitions, it is totally fine that a model does not have a `prediction_key` and it would throw an error in the HFAdapter at runtime. 

Additionally, for different use cases other prediction keys might be expected. E.g., argmax vs softmax outputs are both predictions that we could use in different scenarios. 

I would therefore specify the needed prediction key for the HFAdpater in the config yaml  by referencing the key.

tests/checkpointing/test_checkpoint_conversion.py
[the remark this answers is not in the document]
le1nux: I think it makes it more practical when you also add the type arguments to the fixture arguments. This way autocomplete for instance works and static code analysis such as type checks also work.

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        for key, value in d.items():
+            if isinstance(value, PosixPath):
+                d[key] = str(value)
+            elif isinstance(value, dict):
+                self._convert_posixpath_to_str(value)
+            elif isinstance(value, list):
+                for item in value:
+                    if isinstance(item, PosixPath):
+                        d[key] = str(item)
```
le1nux, line 44: I think this is a bug. 
here we replace a list of Posixpaths and posibly other items to a single posixpath string. Is this wanted behavior?

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+                    elif isinstance(item, dict):
+                        self._convert_posixpath_to_str(item)
+                    else:
+                        d[key] = item
```
le1nux, line 48: same issue as above

src/modalities/models/huggingface_adapters/hf_adapter.py
le1nux: We should cover this method with unit testing.

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class HFModelAdapter(PreTrainedModel):
+    config_class = HFModelAdapterConfig
+
+    def __init__(self, config: HFModelAdapterConfig, *inputs, **kwargs):
+        super().__init__(config, *inputs, **kwargs)
+        self.model: NNModel = get_model_from_config(config.config, model_type="checkpointed_model")
+        assert hasattr(self.model, "prediction_key"), "Missing entry model.prediction_key in config"
```
le1nux, line 57: asserts are often deactivated in production code and it is best practice to raise exceptions instead. 

Following the discusion from below, why don't we pass the prediction key to the constructor of HFModelAdapter, instead? We could prevent this "hack" of checking that a model has a prediction key.

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        return {
+            "input_ids": input_ids,
+            "attention_mask": attention_mask,
+        }
+
+
+@dataclass
+class ModalitiesModelOutput(ModelOutput):
+    logits: torch.FloatTensor = None
```
le1nux, line 91:
```suggestion
    logits: Optional[torch.FloatTensor] = None
```

also applies to the other attribtes below

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+            raise NotImplementedError
+        model_input = {"input_ids": input_ids, "attention_mask": attention_mask}
+        model_forward_output: Dict[str, torch.Tensor] = self.model.forward(model_input)
+        if return_dict:
+            return ModalitiesModelOutput(**model_forward_output)
+        else:
+            return model_forward_output[self.model.prediction_key]
+
+    def prepare_inputs_for_generation(
+        self, input_ids: torch.LongTensor, attention_mask=None, **kwargs
```
le1nux, line 77: type missing for `attention_mask``

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,93 @@
...
+        attention_mask: Optional[torch.Tensor] = None,
+        return_dict: Optional[bool] = False,
+        output_attentions: Optional[bool] = False,
+        output_hidden_states: Optional[bool] = False,
+    ):
+        if output_attentions or output_hidden_states:
+            raise NotImplementedError
+        model_input = {"input_ids": input_ids, "attention_mask": attention_mask}
+        model_forward_output: Dict[str, torch.Tensor] = self.model.forward(model_input)
+        if return_dict:
```
le1nux, line 71: not sure if this is a bug, but if the return_dict is True we return `ModalitiesModelOutput` and the dict otherwise. Shouldn't it be the other way around?

src/modalities/models/utils.py
```diff
@@ -0,0 +1,24 @@
+from pydantic import BaseModel
+
+from modalities.config.component_factory import ComponentFactory
+from modalities.config.pydanctic_if_types import PydanticPytorchModuleType
+from modalities.registry.components import COMPONENTS
+from modalities.registry.registry import Registry
+    
+    
+def get_model_from_config(config: dict, model_type: str):
```
le1nux, line 9: We should make `model_type` an Enum

tests/checkpointing/test_checkpoint_conversion.py
```diff
@@ -0,0 +1,101 @@
+import os
+from pathlib import Path
```
le1nux, line 2: type hints are missing basically everywhere in the tests. That makes it difficult to verify the correctness of the tests.

tests/checkpointing/test_checkpoint_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_models_before_and_after_conversion_are_equal(
+    pytorch_model,
+    hf_model,
+    hf_model_from_checkpoint,
+):
+    for p1, p2, p3 in zip(hf_model.parameters(), pytorch_model.parameters(), hf_model_from_checkpoint.parameters()):
+        assert torch.equal(p1, p2)
+        assert torch.equal(p1, p3)
```
le1nux, line 101: Another test would be great that the outputs  of  `pytorch_model`, `hf_model` and `hf_model_from_checkpoint` are the same for a given random input batch.

le1nux (commented):
The PR is almost in a mergable state. I found one potential bug with the PosixPath to string conversion. And some minor things that I think should be addressed before merging. Other than that it's in a great state and we should be able to merge after including the requested changes and another person has reviewed it :-)

tests/checkpointing/test_checkpoint_conversion.py
rrutmann: We had this test before, but isn't this implied if all parameters are the same?

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        return_dict: Optional[bool] = False,
+        output_attentions: Optional[bool] = False,
+        output_hidden_states: Optional[bool] = False,
+    ):
+        if output_attentions or output_hidden_states:
```
mali-git, line 66: Please add a comment describing this case.

src/modalities/models/huggingface_adapters/hf_adapter.py
mali-git: This is also not clear to me. 
@rrutmann, can you please explain this?

src/modalities/models/utils.py
```diff
@@ -0,0 +1,28 @@
...
+from modalities.config.component_factory import ComponentFactory
+from modalities.config.pydanctic_if_types import PydanticPytorchModuleType
+from modalities.registry.components import COMPONENTS
+from modalities.registry.registry import Registry
+
+class ModelTypeEnum(Enum):
+    MODEL = "model"
+    CHECKPOINTED_MODEL = "checkpointed_model"
+    
+def get_model_from_config(config: dict, model_type: ModelTypeEnum):
```
mali-git, line 13: `dict `is not a type annotation. Please use `Dict` from typing and the type annotations for key an value.

tests/models/test_hf_adapter.py
```diff
@@ -0,0 +1,32 @@
...
+
+from modalities.models.huggingface_adapters.hf_adapter import HFModelAdapterConfig
+
+
+@pytest.fixture()
+def hf_model_adapter_config() -> HFModelAdapterConfig:
+    return HFModelAdapterConfig(config={})
+
+
+def test___convert_posixpath_to_str(hf_model_adapter_config: HFModelAdapterConfig):
```
mali-git, line 13: To many underscore characters

mali-git (changes requested):

PR #168, opened 2024-06-30
Draft: Feat/initialization component

le1nux:
# What does this PR do?
This PR introduces the components for weight initialisation and is based on PR #161. 
 In PR #161 the differenct initialization methods `plain`, `scaled` and `scaled_embed` (see https://arxiv.org/abs/2312.16903) were implemented and added to the abstract `NNModel` class. 
Due to some design concerns (e.g., some GPT2 internals were called from the parent), we decided to introduce a weight initialisation component that modifies the model weights in place. 

## General changes
* Components and factories for  `plain`, `scaled` and `scaled_embed` initialisation.

## Breaking Changes
* The raw model (i.e., the model with random weights) must be initialised with a weight initialiser, as shown [here](https://github.com/Modalities/modalities/blob/7d26675051b918c3a2b98f32f50cb3ca8ef97d6f/config_files/training/config_lorem_ipsum.yaml#L181).

- [x] I have merged the target branch into this feature branch


src/modalities/nn/weight_init/weight_init_factory.py
```diff
@@ -0,0 +1,97 @@
...
+            if hidden_dim is None:
+                raise ValueError("ERROR! weight_init.std = auto not implemented")
+
+            std = math.sqrt(2 / (5 * hidden_dim))
+
+        initialization = ModulewiseNormalInitialization(mean=mean, std=std)
+        return initialization
+
+    @staticmethod
+    def get_named_parameterwise_normal_initialization(
```
le1nux, line 43: currently not used, can be removed

config_files/training/config_lorem_ipsum.yaml
```diff
@@ -178,7 +178,25 @@ wrapped_model:
...
+    weight_initializer:
+      component_key: weight_initialization
+      variant_key: composed
+      config:
+        model_type: gpt2
+        weight_init_type: scaled_embed
+        mean: 0.0
+        plain_std: 0.02
+        hidden_dim: ${model_raw.config.n_embd}
+        num_layers: ${model_raw.config.n_layer}
```
le1nux, line 197: @flxst, are these the expected references?

src/modalities/nn/weight_init/weight_init.py
```diff
@@ -0,0 +1,56 @@
...
+    def _init_weights_impl(self, module: nn.Module):
+        if isinstance(module, nn.Linear):
+            nn.init.normal_(module.weight, mean=self.mean, std=self.std)
+            if module.bias is not None:
+                nn.init.zeros_(module.bias)
+        elif isinstance(module, nn.Embedding):
+            nn.init.normal_(module.weight, mean=self.mean, std=self.std)
+        else:
+            pass
+            # TODO need to check if we want raise an exception here
```
le1nux, line 39: TODO

src/modalities/models/coca/coca_model.py
```diff
@@ -123,14 +121,6 @@ def __init__(
             attention_config=text_decoder_config.attention_config,
         )
 
-        # init all weights
-        assert weight_init.type in ["plain", "scaled"], f"ERROR! weight_init.type = {weight_init.type} not implemented."
-        self.initialize_weights(
```
le1nux, line 128: TODO: check how we want to integrate coca into this

src/modalities/nn/weight_init/weight_init.py
```diff
@@ -0,0 +1,56 @@
...
+
+        Args:
+            mean (float): mean of the normal distribution
+            std (float): standard deviation of the normal distribution
+        """
+        self.mean = mean
+        self.std = std
+
+    def _init_weights_impl(self, module: nn.Module):
+        if isinstance(module, nn.Linear):
```
le1nux, line 31: Need to check how we can support SwiGLU here, with their different layers W, V, W2. 
https://arxiv.org/pdf/2002.05202

see here https://github.com/Modalities/modalities/blob/ec520f4dff20fdd3b68b79be7666b2fbb0e89f42/src/modalities/models/gpt2/gpt2_model.py#L395

le1nux: Also, the manual swiGLU implementation in main, has a linear layer called `c_proj` which is `V` in the original SwiGLU paper. So there is a naming conflict and we would apply `scaled init` to it, instead of `out_proj`, which corresponds to `W_2` in the paper and is the one that actually should receive `scaled_init`.

config_files/training/config_lorem_ipsum.yaml
flxst: Yes, looks good.

src/modalities/nn/weight_init/weight_init.py
flxst: We should probably rename the layers in our manual SwiGLU implementation in main, such that `out_proj = c_proj = W_2`, plus different names for `V` and `W`.

src/modalities/nn/weight_init/low_level_weight_init_factory.py
```diff
@@ -0,0 +1,121 @@
...
+class PlainWeightInitializationConfig(BaseModel):
+    mean: float
+    std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
+    hidden_dim: Optional[int] = None
+
+    @model_validator(mode="after")
+    def check_std_and_hidden_dim(self):
+        if self.std == "auto" and self.hidden_dim is None:
+            raise ValueError("hidden_dim must be specified when std is 'auto'")
+        return self
```
flxst, line 28: `self.std` can be `auto` also for initialization types other than `plain`. So if we want to check this here (we seem to check it in 3 different places in total: `ComposedWeightInitializationConfig`, `PlainWeightInitializationConfig `, `LowLevelInitializationFactory`), we need to do it for `scaled` and `scaled_embed` as well.

src/modalities/nn/weight_init/low_level_weight_init_factory.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Args:
+            mean (float): mean of the normal distribution
+            std (float): standard deviation of the normal distribution
+            hidden_dim (Optional[int], optional): hidden dimension of the attention layer. Defaults to None.
+        """
+
+        # auto: choose std automatically
+        if std == "auto":
+            if hidden_dim is None:
```
flxst, line 71: `std` can be `auto` also for initialization types other than `plain`. So if we want to check this here (we seem to check it in 3 different places in total: `ComposedWeightInitializationConfig`, `PlainWeightInitializationConfig `, `LowLevelInitializationFactory`), we need to do it for `scaled` and `scaled_embed` as well.

src/modalities/nn/weight_init/weight_init.py
flxst: Models other than GPT-2 will have modules that are neither `nn.Linear` nor `nn.Embedding`. For instance, CoCa contains `nn.Conv2d`, see [here](https://github.com/Modalities/modalities/blob/bb4cd363a619792c8d00ccb78416893d0693b354/src/modalities/models/vision_transformer/vision_transformer_model.py#L40). However, `nn.Conv2d` also has the attributes `weight` and `bias` (see [here](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html)). So a pragmatic, first solution could be something like this: We check in the `else` part whether a module has attributes `weight` and (potentially) `bias`. If it does, we use the same initialization as for `nn.Linear`. If it contains other attributes, we raise an Exception.

src/modalities/registry/components.py
```diff
@@ -106,7 +118,41 @@ class ComponentEntity:
...
+        "weight_initialization",
+        "wrapper",
+        LowLevelInitializationFactory.get_weight_initializer_wrapper,
+        WeightInitializerWrapperConfig,
+    ),
+    ComponentEntity(
+        "weight_initialization",
+        "plain",
+        LowLevelInitializationFactory.get_plain_initialization,
+        PlainWeightInitializationConfig,
```
flxst, line 142:
```suggestion
        PlainInitializationConfig,
```
class name would have to be changed too ofc

src/modalities/registry/components.py
```diff
@@ -106,7 +118,41 @@ class ComponentEntity:
...
+    ),
+    ComponentEntity(
+        "weight_initialization",
+        "scaled_weight",
+        LowLevelInitializationFactory.get_scaled_initialization,
+        ScaledWeightInitializationConfig,
```
flxst, line 148:
```suggestion
        ScaledInitializationConfig,
```
class name would have to be changed too ofc

src/modalities/nn/weight_init/low_level_weight_init_factory.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class ScaledEmbedInitializationConfig(BaseModel):
+    mean: float
+    parameter_name_suffixes: List[str]  # here we filter for the parameter names, e.g., "c_proj.weight"
+
+
+class LowLevelInitializationFactory:
+    @staticmethod
+    def get_weight_initializer_wrapper(weight_initializers: List[WeightInitializationIF]) -> WeightInitializationIF:
```
flxst, line 51: This function is used only in `HighLevelWeightInitializationFactory`, and I think it would make more sense to move it there. One might even just get rid of it and use the `WeightInitializerWrapper` class directly, as a simplification.

src/modalities/registry/components.py
```diff
@@ -106,7 +118,41 @@ class ComponentEntity:
...
+    ComponentEntity(
+        "model", "weight_initialized", ModelFactory.get_weight_initalized_model, WeightInitializedModelConfig
+    ),
     ComponentEntity("model", "coca", CoCa, CoCaConfig),
+    # weight initializers
+    ComponentEntity(
+        "weight_initialization",
+        "composed",
+        HighLevelWeightInitializationFactory.get_composed_weight_init,
+        ComposedWeightInitializationConfig,
```
flxst, line 130: I think it is a bit suboptimal to have the naming `LowLevel` vs. `HighLevel`, but then also `Composed`. Maybe we could somehow use only `LowLevel` vs. `HighLevel` OR e.g. `Single` vs. `Composed`.

flxst (changes requested):
Looks good to me generally, although I am a bit concerned about the complexity of this whole implementation. Added a few comments. The ones that contain something like "std can be auto also for initialization types other than plain" are the most important ones, I think this should definitely be fixed.

src/modalities/models/coca/coca_model.py
flxst: [Here](https://github.com/Modalities/modalities/blob/bb4cd363a619792c8d00ccb78416893d0693b354/src/modalities/nn/weight_init/high_level_weight_init_factory.py#L30) we state that only the `plain` initialization can be used for CoCa. I think that is a good starting point. There are just two things missing in my opinion: 
1. We need to exclude the use of `std = 'auto'` for CoCa (the hidden dimension `d` used in `'auto'` mode might not be well-defined as the text and multimodal encoder can in general have different `d`)
2. The unit tests need to be updated.

src/modalities/nn/weight_init/weight_init.py
le1nux: handled here: https://github.com/Modalities/modalities/pull/168/commits/48dced93e783090917fd3a9ba82046064f09d0a8

src/modalities/nn/weight_init/low_level_weight_init_factory.py
```diff
@@ -0,0 +1,121 @@
...
+                raise ValueError("ERROR! weight_init.std = auto not implemented")
+            # as per  https://arxiv.org/abs/2312.16903
+            std = math.sqrt(2 / (5 * hidden_dim))
+
+        initialization = ModulewiseNormalInitialization(mean=mean, std=std)
+        return initialization
+
+    @staticmethod
+    def get_scaled_initialization(
+        mean: float, plain_std: float, num_layers: int, parameter_name_suffixes: List[str]
```
flxst, line 81:
```suggestion
        mean: float, plain_std: float | str, num_layers: int, parameter_name_suffixes: List[str]
```
If `plain_std` is a str, line 97 will be ill-defined and cause an Error. That's why we need the check and the scaling (`std = math.sqrt(2 / (5 * hidden_dim))`) as for the `plain` initialization.

src/modalities/nn/weight_init/low_level_weight_init_factory.py
```diff
@@ -0,0 +1,121 @@
...
+
+class NamedParameterwiseNormalInitializationConfig(BaseModel):
+    mean: float
+    std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
+    parameter_name_suffixes: List[str]  # here we filter for the parameter names, e.g., "c_proj.weight"
+
+
+class ScaledWeightInitializationConfig(BaseModel):
+    mean: float
+    plain_std: Annotated[float, Field(strict=True, ge=0.0)]
```
flxst, line 39:
```suggestion
    plain_std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
```
see other comment

flxst (commented):
I think this is another thing that needs to be fixed.

src/modalities/nn/weight_init/weight_init.py
le1nux: It is implemented here now: 
https://github.com/Modalities/modalities/blob/b8e42c10dba4b61d97a06e21a13ad679ef6cc1cd/src/modalities/nn/weight_init/parameter_name_regex_filters.py#L71-L72

src/modalities/models/coca/coca_model.py
le1nux: Regarding 1., I think we don't have to "exclude" `std='auto'` because, as you said, it is not applicable and one wouldn't specify `hidden_dim` in the config in the first place. 

Yes, unit tests need to be updated.

src/modalities/nn/weight_init/weight_init.py
le1nux: see issue #170. We should fix this when we merge main into the branch of PR #145, since the SwiGLU reimplementation is only available in the latest main version.

config_files/training/config_lorem_ipsum.yaml
```diff
@@ -178,7 +178,25 @@ wrapped_model:
...
+      instance_key: model_raw
+      pass_type: BY_REFERENCE
+    weight_initializer:
+      component_key: weight_initialization
+      variant_key: composed
+      config:
+        model_type: gpt2
+        weight_init_type: scaled_embed
+        mean: 0.0
+        plain_std: 0.02
```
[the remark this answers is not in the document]
le1nux: https://github.com/Modalities/modalities/commit/054d6a2e244832792bf859785642f070ee34fcc0 and https://github.com/Modalities/modalities/commit/602336c9d511b519bfb9a03fac32ef6499e713d9

src/modalities/nn/weight_init/high_level_weight_init_factory.py
```diff
@@ -0,0 +1,138 @@
...
+    },
+}
+
+
+class ComposedWeightInitializationConfig(BaseModel):
+    model_type: SupportWeightInitModels
+    weight_init_type: WeightInitTypes
+
+    mean: float
+    plain_std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
```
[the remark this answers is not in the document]
le1nux: done

src/modalities/nn/weight_init/high_level_weight_init_factory.py
```diff
@@ -0,0 +1,138 @@
...
+    hidden_dim: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
+    num_layers: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
+
+    @model_validator(mode="after")
+    def check_values(self):
+        # in case of the plain initialization with "auto", we need to specify the hidden_dim
```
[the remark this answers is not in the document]
le1nux: done

src/modalities/nn/weight_init/low_level_weight_init_factory.py
le1nux: we now pass the std calculated from `plain` (i.e., working_std) to `scaled`.
See here: https://github.com/Modalities/modalities/blob/824da6ad8fed1fcf0e328d7bffc53ed4c3ffc072/src/modalities/nn/weight_init/high_level_weight_init_factory.py#L106

Therfore only `PlainInitializationConfig` has to allow the "auto" option and the other configs (i.e., `ScaledInitializationConfig` and `ScaledEmbedInitializationConfig`) accept only `float` for `std`

src/modalities/nn/weight_init/low_level_weight_init_factory.py
le1nux: see [comment](https://github.com/Modalities/modalities/pull/168#discussion_r1663192270)

src/modalities/nn/weight_init/low_level_weight_init_factory.py
le1nux: done

src/modalities/nn/weight_init/low_level_weight_init_factory.py
flxst: Ok, I guess that works as well, as long as the `weight_initialization` component with `variant_key = composed` is always used in the config, not the other `variant_key`s, right?

src/modalities/nn/weight_init/low_level_weight_init_factory.py
flxst: resolved by https://github.com/Modalities/modalities/pull/168#discussion_r1663192270

src/modalities/nn/weight_init/low_level_weight_init_factory.py
flxst: resolved by https://github.com/Modalities/modalities/pull/168#discussion_r1663192270

config_files/training/config_example_coca.yaml
```diff
@@ -170,7 +170,23 @@ wrapped_model:
...
+      instance_key: model_raw
+      pass_type: BY_REFERENCE
+    weight_initializer:
+      component_key: weight_initialization
+      variant_key: composed
+      config:
+        model_type: coca
+        weight_init_type: plain
+        mean: 0.0
+        plain_std: 0.02
```
flxst, line 187:
```suggestion
        std: 0.02
```

config_files/training/config_lorem_ipsum.yaml
```diff
@@ -178,7 +178,25 @@ wrapped_model:
...
+      pass_type: BY_REFERENCE
+    weight_initializer:
+      component_key: weight_initialization
+      variant_key: composed
+      config:
+        model_type: gpt2
+        weight_init_type: scaled_embed
+        mean: 0.0
+        std: 0.02
+        hidden_dim: ${model_raw.config.n_embd}
```
flxst, line 196: Since `std != 'auto'`, the `hidden_dim` is not used and should not be specified. See also other comment.

src/modalities/nn/weight_init/high_level_weight_init_factory.py
```diff
@@ -0,0 +1,132 @@
...
+    mean: float
+    std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
+    hidden_dim: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
+    num_layers: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
+
+    @model_validator(mode="after")
+    def _check_values(self):
+        # in case of initialization with "auto", we need to specify the hidden_dim
+        if self.std == "auto" and self.hidden_dim is None:
+            raise ValueError("hidden_dim must be specified when std is 'auto'")
```
flxst, line 33:
```suggestion
            raise ValueError("hidden_dim must be specified when std is 'auto'")
            
        if self.std != "auto" and self.hidden_dim is not None:
            raise ValueError("hidden_dim must not be specified when std is a number")
```

config_files/training/config_example_coca.yaml
flxst: Fixed in [0e29447](https://github.com/Modalities/modalities/pull/168/commits/0e29447002713b84d5bf6c0057b33b964e96cd82)

src/modalities/registry/components.py
le1nux: yeah I was also unsatisfied with the naming. This should improve it: 
https://github.com/Modalities/modalities/pull/168/commits/0e29447002713b84d5bf6c0057b33b964e96cd82

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -0,0 +1,122 @@
...
+            for bias_regex in bias_regexes:
+                if re.fullmatch(bias_regex, parameter_name):
+                    nn.init.zeros_(p)
+
+
+class InitializationRoutines:
+    @staticmethod
+    def get_plain_initialization(
+        mean: float, std: float | str, parameter_name_regexes: List[str], hidden_dim: Optional[int] = None
+    ) -> NamedParameterwiseNormalInitialization:
```
flxst, line 59:
```suggestion
    ) -> ModelInitializationIF
```
To specify the abstract base class instead of the more specific inherited class here is in accordance with what is done for all related methods (`get_scaled_initialization`, `get_scaled_embed_initialization`, `get_composed_model_initializer`, `get_model_initializer_wrapper`).

src/modalities/nn/weight_init/low_level_weight_init_factory.py
le1nux: Yes, for the other variant key `scaled`, we would have to specify `std` in the config as float. I think this is not a limitation, because anyone can still use the `composed` one.

config_files/training/config_lorem_ipsum.yaml
le1nux: added pydantic checks for this case:
https://github.com/Modalities/modalities/commit/eea99bc1ba86923c8e9fbfa410c8bc502e8598f4

src/modalities/nn/weight_init/low_level_weight_init_factory.py
flxst: That's fine, I just wonder whether we should allow people to use the variant keys `plain`, `scaled`, `scaled_embed` (and `wrapper`?) in a config file at all. You can specify everything you want using the variant key `composed`, so why not make it the only one "available" for the sake of simplicity?

src/modalities/nn/weight_init/high_level_weight_init_factory.py
le1nux: added pydantic checks for this case:
https://github.com/Modalities/modalities/commit/eea99bc1ba86923c8e9fbfa410c8bc502e8598f4

src/modalities/nn/model_initialization/initialization_routines.py
le1nux: Generally true, but we need the concrete class in `ComposedInitializationRoutines.get_composed_model_initializer(...)`, as we need to access `plain_init.std`. Otherwise, there is no typing support. 

https://github.com/Modalities/modalities/blob/eea99bc1ba86923c8e9fbfa410c8bc502e8598f4/src/modalities/nn/model_initialization/composed_initialization.py#L116-L120

flxst (approved):

PR #206, opened 2024-08-04
Fsdp 2.0 integration

le1nux:
# What does this PR do?

This PR ..

## General Changes
* ..

## Breaking Changes
* .. 

## TODOs
 - [ ] Need to check if we want to use tied weights for the embedding and LM head and how this would be parallelized with FSDP and tensor parallelism 


src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -18,7 +18,9 @@ def __init__(self, wrapped_model: FSDP, max_norm: float, norm_type=GradientClipp
         self.norm_type = norm_type
 
     def clip_gradients(self) -> torch.Tensor:
-        gradient_norm_score = self.wrapped_model.clip_grad_norm_(max_norm=self.max_norm, norm_type=self.norm_type.value)
+        gradient_norm_score = torch.nn.utils.clip_grad_norm_(
+            self.wrapped_model.parameters(), max_norm=self.max_norm, norm_type=self.norm_type.value, foreach=True
```
le1nux, line 22: this would only calculate the gradient norm w.r.t. to materialized parmeters on each rankj and would be a bug therefore!

see: https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.clip_grad_norm_

```
if at least some FSDP instance uses a sharded strategy (i.e. one other than NO_SHARD), then you should use this method instead of [torch.nn.utils.clip_grad_norm_()](https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html#torch.nn.utils.clip_grad_norm_) since this method handles the fact that gradients are sharded across ranks.
```

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -131,6 +145,24 @@ def _filter_params_for_weight_decay_group(
     ]
 
 
+def _assert_parameters_belong_to_single_group(
```
le1nux, line 148: @flxst  do we need this check?

PR #211, opened 2024-08-06
Fix Getting Started Example

flxst:
# What does this PR do?

This PR
* fixes the getting started example training config
* adds a shell script for automated testing of the getting started example, which is included in (multi-gpu) testing

Additionally, it
* updates the PR template with an additional check

## General Changes
none except for the above

## Breaking Changes
none


examples/getting_started/run_getting_started_example.sh
```diff
@@ -0,0 +1,40 @@
...
+
+first_character=${1:0:1}
+if [[ $first_character =~ [^0-7] ]]   # if the first character is not an integer 0-7
+    then
+        echo "First character of specified argument needs to be an integer, e.g. bash run_getting_started_example.sh 0,1,2,3,4,5,6,7"
+        exit
+fi
+
+last_character=${1:0-1}
+if [[ $last_character =~ [^0-7] ]]   # if the first character is not an integer 0-7
```
le1nux, line 25:
```suggestion
if [[ $last_character =~ [^0-7] ]]   # if the last character is not an integer 0-7
```

Also, shouldn't we check for all of them, rather than only the first and last?

tests/tests.py
```diff
@@ -1,11 +1,28 @@
...
+import os
 import subprocess
-from os.path import isfile
+from os.path import isdir, isfile, join
 from pathlib import Path
 
 _ROOT_DIR = Path(__file__).parents[1]
 
 
+def clear_getting_started_example_output_directory(output_directory):
```
le1nux, line 10: how about `tmpdir`? Then we wouldn't need this check right?

tests/tests.py
```diff
@@ -41,27 +58,46 @@ def main(cpu: bool = False, single_gpu: bool = False, multi_gpu: bool = False, d
...
+        ), f"ERROR! {run_getting_started_example_script} does not exist."
+        command_getting_started_example = (
+            f"cd {run_getting_started_example_directory}; bash run_getting_started_example.sh {devices[0]},{devices[1]}"
+        )
+        print(command_getting_started_example)
+        subprocess.run(command_getting_started_example, shell=True, capture_output=False, text=True)
+
+        output_directory = join(run_getting_started_example_directory, "data", "mem_map")
+        clear_getting_started_example_output_directory(output_directory)
+
```
le1nux, line 98: What about a check that the files have been created including idx, pbin and checkpoint files?

le1nux (changes requested):
Looks good! Also nice idea to run unit testing on the examples as well! 👍 
Left a few minor comments.

examples/getting_started/run_getting_started_example.sh
flxst: Actually, I found another solution now, implemented in efc1629c0c4ef29728dc9cec59e08763274ddf78: The number of GPUs is simply restricted to 2, and the devices are specified like so: `bash run_getting_started_example.sh 0 1`. This allows for simple yet sufficient checking of the input arguments and avoids the aforementioned problems.

tests/tests.py
flxst: I think it would be difficult to use `tmpdir`, since the getting started example is a "real-world" training run where the output file paths are explicitly specified in the config file.

le1nux: true we cannot do that at the moment. In the future, we could define a component, which provides you with a tmp path.

PR #215, opened 2024-08-09
Compute MFU

flxst:
# What does this PR do?

This PR introduces the computation of MFU

## General Changes
none

## Breaking Changes
none

### DIFF
--- a/src/modalities/trainer.py
+++ b/src/modalities/trainer.py
@@ -16,7 +16,14 @@ from modalities.loss_functions import Loss
 from modalities.models.model import model_predict_batch
 from modalities.running_env.fsdp.reducer import Reducer
 from modalities.training.gradient_clipping.gradient_clipper import GradientClipperIF
-from modalities.util import Aggregator, TimeRecorder, print_rank_0
+from modalities.util import (
+    Aggregator,
+    TimeRecorder,
+    compute_mfu,
+    get_theoretical_gpu_peak_performance,
+    get_total_number_of_trainable_parameters,
+    print_rank_0,
+)
 
 
 class ThroughputAggregationKeys(Enum):
@@ -87,7 +94,11 @@ class Trainer:
         model.train()
         cumulated_losses = self._reset_tracked_losses()
 
+        # throughput & MFU
         thoughput_aggregator = Aggregator[ThroughputAggregationKeys]()
+        theoretical_gpu_peak_performance = get_theoretical_gpu_peak_performance()
+        num_params = get_total_number_of_trainable_parameters(model)
+        theoretical_flops_per_token = 6 * num_params  # approximation
 
         device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
@@ -183,12 +194,14 @@ class Trainer:
                 }
                 gradient_norm_scores = []
 
+                mfu = compute_mfu(synced_num_samples, theoretical_flops_per_token, theoretical_gpu_peak_performance)
                 training_metrics = EvaluationResultBatch(
                     losses=losses,
                     metrics=metrics,
                     # TODO: hardcoded metric key
                     throughput_metrics={
                         "train samples/s": synced_num_samples_per_second,
+                        "train mfu": mfu,
                         "lr mean": torch.tensor(scheduler.get_last_lr()).mean(),
                     },
                     dataloader_tag=train_loader.dataloader_tag,
--- a/src/modalities/trainer.py
+++ b/src/modalities/trainer.py
@@ -90,7 +90,7 @@ class Trainer:
 
         # throughput & MFU
         thoughput_aggregator = Aggregator[ThroughputAggregationKeys]()
-        theoretical_gpu_peak_performance = get_theoretical_gpu_peak_performance(world_size=dist.get_world_size())
+        theoretical_gpu_peak_performance = get_theoretical_gpu_peak_performance(model, world_size=dist.get_world_size())
         theoretical_flops_per_token, sequence_length = get_theoretical_flops_per_token(model)
 
         device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
--- a/src/modalities/trainer.py
+++ b/src/modalities/trainer.py
@@ -16,14 +16,8 @@ from modalities.loss_functions import Loss
 from modalities.models.model import model_predict_batch
 from modalities.running_env.fsdp.reducer import Reducer
 from modalities.training.gradient_clipping.gradient_clipper import GradientClipperIF
-from modalities.util import (
-    Aggregator,
-    TimeRecorder,
-    compute_mfu,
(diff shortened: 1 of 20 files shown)


src/modalities/trainer.py
```diff
@@ -87,7 +94,11 @@ def train(
         model.train()
         cumulated_losses = self._reset_tracked_losses()
 
+        # throughput & MFU
         thoughput_aggregator = Aggregator[ThroughputAggregationKeys]()
+        theoretical_gpu_peak_performance = get_theoretical_gpu_peak_performance()
+        num_params = get_total_number_of_trainable_parameters(model)
```
mali-git, line 100: We need to fix before the computations of the number of parameters for hybrid sharding.

src/modalities/util.py
```diff
@@ -180,3 +180,34 @@ def get_module_class_from_name(module: torch.nn.Module, name: str) -> Type[torch
...
             if module_class is not None:
                 return module_class
+
+
+def get_theoretical_gpu_peak_performance() -> Optional[Number]:
+    """
+    returns theoretical gpu peak performance in units FLOPs / s for given gpu type
+    """
+    device_name = torch.cuda.get_device_name()
+    if device_name.startswith("NVIDIA A100"):
```
mali-git, line 190: It might be better to define constants and import them here.

src/modalities/util.py
```diff
@@ -180,3 +180,34 @@ def get_module_class_from_name(module: torch.nn.Module, name: str) -> Type[torch
...
+        return 312e12  # TODO: double-check (also floating point precision types)
+    elif device_name.startswith("NVIDIA H100"):
```
mali-git, line 192: Please see @[mali-git](https://github.com/mali-git) mali-git Pending

src/modalities/util.py
```diff
@@ -180,3 +180,34 @@ def get_module_class_from_name(module: torch.nn.Module, name: str) -> Type[torch
...
+        return 989e12  # TODO: double-check (also floating point precision types)
```
mali-git, line 193: Please see @[mali-git](https://github.com/mali-git) mali-git Pending

src/modalities/trainer.py
flxst: done in #216

src/modalities/util.py
flxst: see reply above

fromm-m (approved):
looks good to me 👍

src/modalities/utils/mfu.py
```diff
@@ -0,0 +1,94 @@
...
+    "A100": 312e12,
+    "H100": 989e12,
+}
+
+
+def _get_theoretical_gpu_peak_performance_single(precision: torch.dtype, gpu_type: str) -> Optional[Number]:
+    """
+    returns theoretical gpu peak performance for #GPU=1 in units FLOPs / s for given gpu type
+    """
+    if precision in [torch.float16, torch.bfloat16] and gpu_type in PEAK_PERFORMANCE.keys():
```
mali-git, line 25:
```suggestion
    if precision in [torch.float16, torch.bfloat16] and gpu_type in PEAK_PERFORMANCE:
```

src/modalities/utils/mfu.py
```diff
@@ -0,0 +1,94 @@
...
+        return PEAK_PERFORMANCE[gpu_type]
+    else:
+        return None
+
+
+def get_theoretical_gpu_peak_performance(model: FSDP, world_size: int) -> Optional[Number]:
+    """
+    returns theoretical gpu peak performance for #GPU=world_size in units FLOPs / s for given gpu type
+    """
+    if torch.cuda.is_available() and torch.cuda.device_count() > 0:  # necessary for cpu-only tests
```
mali-git, line 35: If torch.cuda.device_count() > 0 is only required for CPU related tests, wouldn't it be prefferable to mock this functions for CPU tests?

src/modalities/utils/mfu.py
```diff
@@ -0,0 +1,94 @@
...
+        precision = model.mixed_precision.param_dtype
+        if model.mixed_precision.reduce_dtype != precision or model.mixed_precision.buffer_dtype != precision:
+            warnings.warn("could not get theoretical gpu peak performance for given mixed precision type")
+            return None
```
mali-git, line 39:
```suggestion
            warnings.warn(f"Could not get theoretical GPU peak performance for given mixed precision type {precision}.")
```

src/modalities/utils/mfu.py
```diff
@@ -0,0 +1,94 @@
...
+def get_theoretical_flops_per_token(model: FSDP) -> Tuple[Optional[int], Optional[int]]:
+    """
+    compute theoretical_flops_per_token = 6*N + 12*L*T*H
+    see App. B in the PaLM paper (https://arxiv.org/pdf/2204.02311)
+
+    Returns:
+        theoretical_flops_per_token
+        sequence_length (needed to convert samples to tokens in compute_mfu)
+    """
+    if torch.cuda.is_available() and torch.cuda.device_count() > 0:  # necessary for cpu-only tests
```
mali-git, line 67: torch.cuda.device_count() > 0: See comment above.

mali-git (changes requested):

src/modalities/utils/mfu.py
flxst: I guess you meant to improve the warning message, not to replace `return None`? I fixed this (and related warning messages) in e5af3a44d0f63669b0005083975662f1332e5db1

src/modalities/utils/mfu.py
flxst: The function is inherently GPU-related and tested as part of the GPU tests, so I do not see the need to include it in the CPU tests using mocking. The comment "necessary for cpu-only tests" is a bit misleading though: While the `if` statement also makes sure that the CPU tests run through, its main purpose is to avoid the computation and return early & gracefully in the case of no GPUs.

I simply removed the comment in 2708f37c9ca1e0b7d3db64ea52b2370c8d2f7955

src/modalities/utils/mfu.py
flxst: This case is different from the one above. The function here depends only on the model architecture and in principle does not rely on the presence of GPUs. However, the employed function `get_total_number_of_trainable_parameters` requires GPUs. As you mentioned, the `if` statements could be removed if mocking of the `get_total_number_of_trainable_parameters` function was used for CPU tests. 

I changed the comment in https://github.com/Modalities/modalities/commit/2708f37c9ca1e0b7d3db64ea52b2370c8d2f7955 accordingly, and opened an issue: https://github.com/Modalities/modalities/issues/238

src/modalities/utils/mfu.py
mali-git: I see, thanks for the clarification.

PR #239, opened 2024-08-29
Interactive getting started tutorial

le1nux:
# What does this PR do?

This PR adds an interactive jupyter notebook-based Tutorial on how to use Modalities. 
We also added some fixes for warning output etc. 

## General Changes
* metrics are now printed only with a fixed number of decimals in the console during training. They are still logged to Weights and Biases with the full precision. 
*  removed abundant print statements

## Breaking Changes
* .. 


src/modalities/models/model_factory.py
```diff
@@ -64,10 +62,10 @@ def get_fsdp_wrapped_model(
             'FSDPTransformerAutoWrapPolicyFactory` is hardcoded and should be passed in instead.
             Different auto wrap policies may be supported in the future.
         """
-        print(
-            f"Unsharded number of parameters on rank {dist.get_rank()}: "
-            f"{get_local_number_of_trainable_parameters(model)}"
-        )
+        # print(
```
mali-git, line 65: It would be great to print the number of parameters at least on one rank.

src/modalities/models/model_factory.py
```diff
@@ -82,10 +80,10 @@ def get_fsdp_wrapped_model(
             sync_module_states=sync_module_states,
             use_orig_params=True,
         )
-        print(
-            f"Sharded number of parameters on rank {dist.get_rank()}:"
-            f"{get_local_number_of_trainable_parameters(fsdp_model)}"
-        )
+        # print(
```
mali-git, line 83: It would be great to print the number of parameters at least on one rank.

src/modalities/trainer.py
```diff
@@ -260,9 +260,9 @@ def train(
                     metrics=metrics,
                     # TODO: hardcoded metric key
                     throughput_metrics={
-                        "train samples/s": synced_num_samples_per_second,
-                        "train mfu": mfu,
-                        "lr mean": torch.tensor(scheduler.get_last_lr()).mean(),
+                        "train samples/s": ResultItem(synced_num_samples_per_second, 1),
+                        "train mfu": ResultItem(mfu, 2),
+                        "lr mean": ResultItem(torch.tensor(scheduler.get_last_lr()).mean(), 6),
```
mali-git, line 265: Since the learning rate can be arbitrarily small, I would log it in full precision.

mali-git (changes requested):

src/modalities/trainer.py
le1nux: done

PR #268, opened 2024-11-13
fix: clone TrainingProgress when saving list of saved checkpoints

sthoduka:
# What does this PR do?

The list of `saved_step_checkpoints` contains TrainingProgress objects which are *references* to the TrainingProgress objects, which are updated during training. So all elements in the list will be the same, and will correspond to the current progress. Thus, in cases where `k > 0`, only the first k checkpoints are saved, and all remaining checkpoints are created and immediately deleted (since the `checkpoints_to_delete` is the same as the most recently saved one).  

Solution: clone TrainingProgress object when saving the list of saved checkpoints.

## General Changes
* fixed as above
* added an assert in the test to check for this case

## Breaking Changes
none

- [x] I have checked that all tests run through (`python tests/tests.py`) ([here](https://github.com/Modalities/modalities/actions/runs/11820173760))

### DIFF
--- a/src/modalities/checkpointing/checkpoint_saving_strategies.py
+++ b/src/modalities/checkpointing/checkpoint_saving_strategies.py
@@ -1,3 +1,4 @@
+import dataclasses
 from abc import ABC, abstractmethod
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


le1nux (approved):
Good find and thanks for fixing! 
LGTM :)

PR #269, opened 2024-11-21
Large file reader efficiency improvements and byte reading support

le1nux:
# What does this PR do?

This PR makes the LargeFileLinesReader about 50% faster by using mmap instead of file seek operations. 
We can also now configure the encoding used for reading the documents. If encoding is specifically set to None (default it utf-8), we return the document as a byte string not enforcing any encoding. This is especially helful when we e.g., sample from the data and want to create a subset of the dataset. In this case, we can just pass around the bytes representation. 

## General Changes
* ..

## Breaking Changes
* None


flxst (changes requested):
This looks great! I just find a few minor issues that should perhaps be addressed.

src/modalities/__main__.py
```diff
@@ -132,7 +132,7 @@ def data():
     default=None,
     help="output path for index. will use parent directory of src_path if none.",
 )
-def entry_point_data_create_raw_index(src_path: Path, index_path: Path):
+def CMD_entry_point_data_create_raw_index(src_path: Path, index_path: Path):
```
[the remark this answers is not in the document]
le1nux: I had issues with calling the function in the pytest directly. It was always complaining about the click context. As a workaround I added this trivial wrapper.

src/modalities/dataloader/large_file_lines_reader.py
```diff
@@ -72,27 +93,36 @@ def __len__(self) -> int:
...
+
         return self._read_from_raw_file(offset, sample_length_in_bytes)
 
-    def _read_from_raw_file(self, offset: int, sample_length_in_bytes: int) -> str:
+    def _read_from_raw_file(self, offset: int, sample_length_in_bytes: int) -> str | bytes:
         # Reads a specified number of bytes from a raw file starting from a given offset.
-        f = self.raw_data_path.open()
-        f.seek(offset)
-        return f.read(sample_length_in_bytes)
+        # whence = 0 means offse calculated from the beginning of the file
```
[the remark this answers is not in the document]
le1nux: This was legacy. `whence` was the argument of f.seek. I removed the comment.

tests/dataloader/test_large_file_lines_reader.py
```diff
@@ -51,16 +51,40 @@ def generate_data_index_file(data_path: Path, **kwargs):
...
+        assert all([item[-1] == "\n" for item in reader][:-1])
     # content of dummy data contains trailing "\n"-char. Expected amount of samples therefore == amount of lines - 1
     assert len(reader) == 12
     assert all(map(len, reader))
 
 
+def test_large_file_lines_reader_binary_text_equivalence(indexed_dummy_data_path):
+    raw_data_path = indexed_dummy_data_path.raw_data_path
+    reader_binary = LargeFileLinesReader(raw_data_path, use_sample_length_from_index=False, encoding=None)
+    reader_text = LargeFileLinesReader(raw_data_path, use_sample_length_from_index=False, encoding="utf-8")
```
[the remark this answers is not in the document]
le1nux: good point! Improved in cf82e3672d591ac05fe0c4d140d66a00a9c2827f

tests/dataloader/test_large_file_lines_reader.py
```diff
@@ -51,16 +51,40 @@ def generate_data_index_file(data_path: Path, **kwargs):
...
+
+    for item_binary, item_text in zip(reader_binary, reader_text):
+        assert item_binary.decode("utf_8") == item_text
+
+
 def test_large_file_lines_reader_missing_source_data(tmpdir, dummy_data_path):
```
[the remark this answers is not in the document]
le1nux: Removed tmpdir

src/modalities/__main__.py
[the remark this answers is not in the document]
flxst: Ok, not particularly beautiful, but I guess this is not a problem :)

flxst (approved):

le1nux:
@flxst I added `api.py`,  separating the click CMD functions from the actual implementations. This way, we can use the API functions as imports in other projects.

flxst:
> @flxst I added `api.py`, separating the click CMD functions from the actual implementations. This way, we can use the API functions as imports in other projects.

Looks very clean and good :)

flxst (approved):

src/modalities/dataloader/large_file_lines_reader.py
```diff
@@ -72,27 +93,38 @@ def __len__(self) -> int:
         """
         return len(self.index)
 
-    def __getitem__(self, key: int | slice) -> str | list[str]:
```
rrutmann, line 75: Why did you remove support for slices? The signature is now not consistent with BaseReader.__getitem__

src/modalities/dataloader/large_file_lines_reader.py
```diff
@@ -41,6 +55,13 @@ def __init__(self, raw_data_path: Path, index_path: Optional[Path] = None):
         with self.index_path.open("rb") as f:
             self.index = pickle.load(f)
 
+        self.raw_data_fd = self.raw_data_path.open("rb")
+        self.mmapped_data_file = mmap.mmap(self.raw_data_fd.fileno(), 0, access=mmap.ACCESS_READ)
+
+    def close(self):
```
rrutmann, line 61: The close-function is never called in our codebase. What is the purpose of adding it then? Also, if we keep this function, we should add an open-function as well where we open the appropriate files again

src/modalities/dataloader/large_file_lines_reader.py
```diff
@@ -41,6 +55,13 @@ def __init__(self, raw_data_path: Path, index_path: Optional[Path] = None):
         with self.index_path.open("rb") as f:
             self.index = pickle.load(f)
 
+        self.raw_data_fd = self.raw_data_path.open("rb")
+        self.mmapped_data_file = mmap.mmap(self.raw_data_fd.fileno(), 0, access=mmap.ACCESS_READ)
```
rrutmann, line 59: The __init__-method opens these files, but there is no guarantee that close() is actually called. We could use a context manager (__enter__ and __exit__) for better resource management.

`def __enter__(self):
    return self

def __exit__(self, exc_type, exc_value, traceback):
    self.close()
`

tests/dataloader/test_large_file_lines_reader.py
```diff
@@ -51,17 +51,55 @@ def generate_data_index_file(data_path: Path, **kwargs):
     ] == dummy_content_text.split("\n")
 
 
-def test_large_file_lines_reader(indexed_dummy_data_path):
+@pytest.mark.parametrize(
+    "use_sample_length_from_index",
+    [True, False],
+)
+def test_large_file_lines_reader_text(indexed_dummy_data_path, use_sample_length_from_index):
```
rrutmann, line 58: Missing type annotations

tests/dataloader/test_large_file_lines_reader.py
```diff
@@ -51,17 +51,55 @@ def generate_data_index_file(data_path: Path, **kwargs):
...
     assert len(reader) == 12
     assert all(map(len, reader))
 
 
-def test_large_file_lines_reader_missing_source_data(tmpdir, dummy_data_path):
+@pytest.mark.parametrize(
+    "use_sample_length_from_index",
+    [True, False],
+)
+def test_large_file_lines_reader_binary_text_equivalence(indexed_dummy_data_path, use_sample_length_from_index: bool):
```
rrutmann, line 87: Missing type annotation

tests/dataloader/test_large_file_lines_reader.py
```diff
@@ -51,17 +51,55 @@ def generate_data_index_file(data_path: Path, **kwargs):
...
+        raw_data_path, use_sample_length_from_index=use_sample_length_from_index, encoding="utf-8"
+    )
+
+    for item_binary, item_text in zip(reader_binary, reader_text):
+        assert item_binary.decode("utf_8") == item_text
+        # make sure that when we use sample length from index, we do not have a trailing "\n"-char
+        assert use_sample_length_from_index == (not item_text.endswith("\n"))
+
+
+def test_large_file_lines_reader_missing_source_data(dummy_data_path):
```
rrutmann, line 102: Missing type annotation

src/modalities/api.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Raises:
+        ValueError: If the index file already exists.
+    """
+    index_path = LargeFileLinesReader.default_index_path(src_path, index_path)
+    os.makedirs(index_path.parent, exist_ok=True)
+    if index_path.exists():
+        raise ValueError("index already exists. delete it or specify different output folder.")
+
+    print(f"reading raw data from {src_path}")
```
rrutmann, line 40: The code contains multiple print-statements. Shouldn't we use our logging-functionality or the logging-module instead?

src/modalities/api.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Args:
+        src_path (Path): The path to the jsonl-file.
+        index_path (Path): The path to the index file, that will be created.
+
+    Raises:
+        ValueError: If the index file already exists.
+    """
+    index_path = LargeFileLinesReader.default_index_path(src_path, index_path)
+    os.makedirs(index_path.parent, exist_ok=True)
```
rrutmann, line 36: Why not use pathlib consistently with index_path.mkdir(...) ?

rrutmann (changes requested):
Overall the PR is in a very good state. I added some minor suggestions. I would request changes for the missing type annotations and the wrong function signature (LargeFileLinesReader.__getitem__). The other changes would be optional

src/modalities/api.py
le1nux: I think it is fine for now. It is just for the index generation and printed only once.

src/modalities/dataloader/large_file_lines_reader.py
le1nux: We never used slicing and this simplifies the interface. That's why I removed it. If later on we need it for some reason, we can add it back.

src/modalities/dataloader/large_file_lines_reader.py
le1nux: I created the issue  #271  for this.

src/modalities/dataloader/large_file_lines_reader.py
le1nux: I created the issue https://github.com/Modalities/modalities/issues/271 for this.

PR #281, opened 2024-12-12
Bugfix: char-based index and byte-based index mismatch

le1nux:
# What does this PR do?

The char-based index is not always consistent with the byte-based index.

For instance, the first character of the string "ø This is..." is written on disc as two bytes, namely \xc3\xb8, when encoded as utf-8. 
Therefore, the byte-based index has one more byte/char than the char-based index. 

For consistency, we don't consider any char-based indexes anymore and always refer to byte-based indexes. 


fromm-m (approved):
LGTM

flxst (approved):
LGTM

PR #300, opened 2025-02-10
Added sequential sampler

le1nux:
# What does this PR do?

This PR adds a `SequentialSampler` component.
Before, we only had a `ResumableDistributedSampler` which distributes the samples among all ranks (disjoint sets). For very small datasets this `ResumableDistributedSampler` leads to incomplete batches (or even empty batches on some ranks) when scaling up the ranks. 

As a workaround, the `SequentialSampler` can be used which replicates the data on all ranks instead. This way, the same loss is computed  on each rank. Since we always calculate the per-token loss, this has no effect on the total loss. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,4 @@ noteboks/*
 tests/tmp/*
 *wandb_storage*
 .coverage/*
+eurolingua/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
   rev: 23.9.1
   hooks:
   - id: black
-    language_version: python3.11
+    language_version: python3.12
     stages: [pre-commit]
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.0.278
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
   rev: 23.9.1
   hooks:
   - id: black
-    language_version: python3.12
+    language_version: python3.11
     stages: [pre-commit]
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.0.278
--- a/tests/dataloader/samplers/test_sequential_samplers.py
+++ b/tests/dataloader/samplers/test_sequential_samplers.py
@@ -1,3 +1,5 @@
+import pytest 
+
 from torch.utils.data import SequentialSampler, Dataset
 
 class DummyDataset(Dataset):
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -162,3 +162,4 @@ noteboks/*
 tests/tmp/*
 *wandb_storage*
 .coverage/*
+eurolingua/
```
mali-git, line 165: Modalities should be EuroLingua independent.

mali-git (approved):

.gitignore
mali-git, line 165:
```suggestion
```

mali-git:
@le1nux, a test is missing.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
   rev: 23.9.1
   hooks:
   - id: black
-    language_version: python3.11
+    language_version: python3.12
```
flxst, line 15: Why 3.12? Modalites officially supports 3.10 & 3.11, see [here](https://github.com/Modalities/modalities/blob/782a1fd57563927dacd30326a07ba22ce8c09ab7/pyproject.toml#L4)

tests/dataloader/samplers/test_sequential_samplers.py
```diff
@@ -0,0 +1,23 @@
...
+    def __getitem__(self, index):
+        return self.data[index]
+
+
+def test_distributed_setting():
+    num_samples = 10
+    # Ensure each rank gets same samples
+    dataset = DummyDataset(num_samples)
+    world_size = 3
+    samplers = [SequentialSampler(dataset)for _ in range(world_size)]
```
flxst, line 18:
```suggestion
    samplers = [SequentialSampler(dataset) for _ in range(world_size)]
```

tests/dataloader/samplers/test_sequential_samplers.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def __getitem__(self, index):
+        return self.data[index]
+
+
+def test_distributed_setting():
+    num_samples = 10
+    # Ensure each rank gets same samples
+    dataset = DummyDataset(num_samples)
+    world_size = 3
```
flxst, line 17: Shouldn't we parametrize this test and use at least one different combination of `num_samples` and `world_size`?

flxst (changes requested):
It seems there is a tiny bug in the code. Added 2 other comments as well.

tests/dataloader/samplers/test_sequential_samplers.py
mali-git: Done in b1fb9af43003c7a03a0d0fd4d707d3bd6b876754.

tests/dataloader/samplers/test_sequential_samplers.py
mali-git: Done in b1fb9af43003c7a03a0d0fd4d707d3bd6b876754.

PR #306, opened 2025-02-21
Eval modalities2

BlueCrescent:
# What does this PR do?

This PR adds HFTokenizerAdapter for wrapping sentencepiece based tokenizers used in modalities for usage in hf transformers.
Additionally, a fix in HFModelAdapter is included addressing errors caused by internal changes to transformers.

## General Changes
* ..

## Breaking Changes
* .. 


src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -1,22 +1,27 @@
...
+SPIECE_UNDERLINE = "▁"
 
 
 class HFModelAdapterConfig(PretrainedConfig):
     """HFModelAdapterConfig configuration class for the HFModelAdapter."""
 
     model_type = "modalities"
 
-    def __init__(self, **kwargs):
+    def __init__(self, config={}, **kwargs):
```
BlueCrescent, line 24: From Klaudia: Using a mutable default (config={}), which can cause unintended side effects (Maybe use None instead and initialize it conditionally).

src/modalities/models/huggingface_adapters/hf_adapter.py
```diff
@@ -158,3 +164,236 @@ class ModalitiesModelOutput(ModelOutput):
...
+        `unk_token`. Here is an example with `unk_token = "<unk>"` and `unk_token_length = 4`.
+        `self.tokenizer.sp_model.encode("<unk> Hey", out_type = str)[4:]`.
+        """
+        if self.legacy or not text.startswith((SPIECE_UNDERLINE, " ")):
+            return self.sp_model.tokenizer.encode(text, out_type=str)
+
+            # 1. Encode string + prefix ex: "<unk> Hey"
+        tokens = self.sp_model.tokenizer.encode(self.unk_token + text, out_type=str)
+        # 2. Remove self.unk_token from ['<','unk','>', '▁Hey']
+        return tokens[self.unk_token_length :] if len(tokens) >= self.unk_token_length else tokens
```
BlueCrescent, line 365: From Klaudia: The tokenization logic manually adds <unk> before encoding and removes it afterward, assuming consistent tokenization. If tokenization behavior changes (e.g., due to different training conditions), this approach may break.

src/modalities/models/huggingface_adapters/hf_adapter.py
BlueCrescent: A safeguard has been implemented to make errors due to the mutable default argument less likely.

src/modalities/models/huggingface_adapters/hf_adapter.py
ajude2s: This is required since llama2 tokenizer has removed `add_dummy_prefix` option, which automatically adds a leading white space every input, even if it’s unnecessary, which could interfere with tokenization. This new method introduced in Llama2 of using `<unk>` basically allows us to control when spaces should be preserved. The leading spaces are preserved only if it is originally present in the input. 

>  if tokenization behavior changes

did you mean if the tokenizer is trained to interpret `<unk>` differently? The function `unk_token_length` dynamically calculates how many tokens represent the <unk> token when it is encoded. If the `<unk>` token is treated as a single token (e.g., `['<unk>']`), the length will be 1. If it's split into multiple tokens (e.g., `['<', 'unk', '>']`), the length will be 3.
By using this dynamically calculated length, the tokenizer ensures that it removes the correct number of tokens — no matter how the <unk> token is treated internally by the tokenizer.

```
tokens = ['<unk>', '▁Hey']
unk_token_length = 1
# After removing the `<unk>` token (with length 1)
tokens = tokens[1:]  # Result: ['▁Hey']
```

```
tokens = ['<', 'unk', '>', '▁Hey']
unk_token_length = 3
# After removing the `<unk>` token (with length 3)
tokens = tokens[3:]  # Result: ['▁Hey']
```

PR #312, opened 2025-03-03
Minimal FSDP2 imlementation

le1nux:
# What does this PR do?

This PR ..

## General Changes
* ..

## Breaking Changes
* .. 

### DIFF
--- a/src/modalities/running_env/fsdp/device_mesh.py
+++ b/src/modalities/running_env/fsdp/device_mesh.py
@@ -1,0 +1,103 @@
+from typing import Annotated, Optional
+
+from pydantic import BaseModel, Field, model_validator
+from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
+
+from modalities.exceptions import ConfigError
+from modalities.util import print_rank_0
+
+
+class DeviceMeshConfig(BaseModel):
+    # inspired by ParallelDims class in
+    # https://github.com/pytorch/torchtitan/blob/cfc0f4e08dc71685cdcb394464187d2eeedd1a5f/torchtitan/parallelisms/parallel_dims.py#L15
+    device_type: str = "cuda"
+    data_parallel_replicate_degree: Annotated[int, Field(strict=True, ge=-1)]
+    data_parallel_shard_degree: Annotated[int, Field(strict=True, ge=-1)]
+    tensor_parallel_degree: Annotated[int, Field(strict=True, gt=0)]
+    pipeline_parallel_degree: Annotated[int, Field(strict=True, gt=0)]
+    context_parallel_degree: Annotated[int, Field(strict=True, gt=0)]
+    enable_loss_parallel: Optional[bool] = False
+    world_size: Annotated[int, Field(strict=True, gt=0)]
+
+    @model_validator(mode="after")
+    def _validate(self):
+        if (
+            self.tensor_parallel_degree != 1
+            or self.pipeline_parallel_degree != 1
+            or self.context_parallel_degree != 1
+            or self.enable_loss_parallel
+        ):
+            raise ConfigError("Only tensor parallelism, data parallelism and loss parallelism are not supported, yet.")
+
+        for d in (
+            self.data_parallel_replicate_degree,
+            self.context_parallel_degree,
+            self.tensor_parallel_degree,
+            self.pipeline_parallel_degreep,
+        ):
+            if d < 1:
+                raise ConfigError("Parallelism degree must be >= 1, except for data_parallel_shard_degree")
+
+        if not (self.data_parallel_shard_degree == -1 or self.data_parallel_shard_degree >= 1):
+            raise ConfigError("data_parallel_shard_degree must be -1 or >= 1")
+
+        if self.data_parallel_shard_degree == -1:
+            # set the shard degree to the world size divided by the product of all other parallel degrees
+            self.data_parallel_shard_degree = self.world_size // (
+                self.data_parallel_replicate_degree
+                * self.context_parallel_degree
+                * self.tensor_parallel_degree
+                * self.pipeline_parallel_degree
+            )
+        if (
+            self.data_parallel_shard_degree
+            * self.data_parallel_replicate_degree
+            * self.tensor_parallel_degree
+            * self.pipeline_parallel_degree
+            * self.context_parallel_degree
+            != self.world_size
+        ):
+            raise ConfigError(
+                f"Invalid parallel dims: data_parallel_shard_degree({self.data_parallel_shard_degree}) * "
+                f"data_parallel_replicate_degree({self.data_parallel_replicate_degree}) * "
+                f"tensor_parallel_degree({self.tensor_parallel_degree}) *"
+                f"* pipeline_parallel_degree({self.pipeline_parallel_degree}) *"
+                f"context_parallel_degree({self.context_parallel_degree})!= WORLD_SIZE({self.world_size})"
+            )
+        if self.enable_loss_parallel and self.tensor_parallel_degree <= 1:
+            raise ConfigError(f"{self.enable_loss_parallel=} requires tensor_parallel_degree > 1")
+        return self
+
+
+def get_device_mesh(
+    device_type: str,
+    data_parallel_replicate_degree: int,
+    data_parallel_shard_degree: int,
+    tensor_parallel_degree: int,
+    pipeline_parallel_degree: int,
+    context_parallel_degree: int,
+    enable_loss_parallel: bool,
+    world_size: int,
+) -> DeviceMesh:
+    dims = []
+    names = []
+    for dim, name in zip(
+        [
+            pipeline_parallel_degree,
+            data_parallel_replicate_degree,
+            data_parallel_shard_degree,
+            context_parallel_degree,
+            tensor_parallel_degree,
+        ],
+        ["pp", "dp_replicate", "dp_shard", "cp", "tp"],
+        strict=True,
+    ):
+        if dim > 1:
+            dims.append(dim)
+            names.append(name)
+    names = tuple(names)
+    device_mesh = init_device_mesh(device_type, dims, mesh_dim_names=names)
+    print_rank_0(f"{device_mesh=} | {world_size=} | {enable_loss_parallel=}")
+    # TODO: Torch Titan had some more checks here. We need to check if we also need those:
+    # https://github.com/pytorch/torchtitan/blob/b291ad662493b63d25b038a30a915082d3617baf/torchtitan/distributed/parallel_dims.py#L86-L104
+    return device_mesh


src/modalities/checkpointing/fsdp/fsdp_checkpoint_saving.py
```diff
@@ -61,77 +74,114 @@ def __init__(
...
-        entity_file_name = self.CHECKPOINT_STRUCTURE.format(
-            experiment_id=experiment_id,
-            entity=entity_type.value,
-            num_seen_steps=str(num_seen_steps),
-            num_seen_tokens=str(num_seen_tokens),
-            num_target_steps=str(num_target_steps),
-            num_target_tokens=str(num_target_tokens),
-        )
+        if entity_type == CheckpointingEntityType.DISTRIBUTED:
+            full_path = Path(self.checkpoint_path, experiment_id, 'distributed_checkpoint')
```
le1nux, line 84: Is there a reason that we have `distributed_checkpoint` fixed here? Would an folder_name like 
`eid_<experiment_id>-seen_steps_<seen_steps>-seen_tokens_<seen_tokens>-target_steps_<target_steps>-target_tokens_<target_tokens>` be more applicable and consistent with what we had before?

flxst: Yes, good idea, we should change that

le1nux:
fixes #315

le1nux:
closes #325

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+    LR_SCHEDULER = "lr_scheduler"
+
+
+class AppState(Stateful):
+    """
+    This is a useful wrapper for checkpointing the application state (i.e., model, optimizer, lr scheduler).
+    Since this object is compliant with the Stateful protocol, DCP will automatically call
+    state_dict/load_stat_dict as needed in the dcp.save/load APIs.
+
+    Note: We take advantage of this wrapper to hande calling distributed state dict methods on the model
```
mali-git, line 31:
```suggestion
    Note: We take advantage of this wrapper to handle calling distributed state dict methods on the model
```

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+
+        Returns:
+            dict[str, Any]: The state dict of the AppState object.
+        """
+        raise NotImplementedError
+
+
+class ModelStateRetriever(StateRetrieverIF):
+    @staticmethod
+    def get_state_dict(app_state: AppState) -> dict[str, Any]:
```
mali-git, line 155: Add docstring.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+        return get_model_state_dict(model=app_state.model)
+
+    @staticmethod
+    def load_state_dict_(app_state: AppState, state_dict: dict[str, Any]) -> None:
```
mali-git, line 159: Add docstring.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+        set_model_state_dict(model=app_state.model, model_state_dict=state_dict, options=StateDictOptions(strict=False))
+
+
+class OptimizerStateRetriever(StateRetrieverIF):
+    @staticmethod
+    def get_state_dict(app_state: AppState) -> dict[str, Any]:
```
mali-git, line 165: Add docstring.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+        sd = get_optimizer_state_dict(
+            model=app_state.model,
+            optimizers=app_state.optimizer,
+            options=StateDictOptions(flatten_optimizer_state_dict=True),
+        )
+        return sd
+
+    @staticmethod
+    def load_state_dict_(app_state: AppState, state_dict: dict[str, Any]) -> None:
```
mali-git, line 174: Add docstring.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+            model=app_state.model,
+            optimizers=app_state.optimizer,
+            optim_state_dict=state_dict,
+            options=StateDictOptions(flatten_optimizer_state_dict=True),
+        )
+
+
+class LRSchedulerStateRetriever(StateRetrieverIF):
+    @staticmethod
+    def get_state_dict(app_state: AppState) -> dict[str, Any]:
```
mali-git, line 185: Add docstring.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+        return app_state.lr_scheduler.state_dict()
+
+    @staticmethod
+    def load_state_dict_(app_state: AppState, state_dict: dict[str, Any]) -> None:
```
mali-git, line 189: Add docstring.

src/modalities/checkpointing/stateful/app_state_factory.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class AppStateFactory:
+    """Factory class to create AppState objects."""
+
+    @staticmethod
+    def get_raw_app_state(
+        model: nn.Module, optimizer: Optimizer, lr_scheduler: Optional[LRScheduler] = None
+    ) -> AppState:
+        """Creates a new (non-checkpoint loaded) AppState object from an instanitated
```
mali-git, line 20:
```suggestion
        """Creates a new (non-checkpoint loaded) AppState object from an instantiated
```

src/modalities/checkpointing/checkpoint_loading.py
```diff
@@ -4,8 +4,30 @@
...
+        Raises:
+            NotImplementedError: This abstract method is not implemented and should be overridden in a subclass.
+
+        Returns:
+            AppState: The application state with the loaded checkpoint.
+        """
+        raise NotImplementedError
+
+
+class FSDP1CheckpointLoadingIF(ABC):
```
mali-git, line 30: Does the interface contain FSDP1-specific logic for loading checkpoints? If not, I would opt keeping the name "CheckpointLoadingIF" or "NoneDistributedCheckpointLoadingIF".

src/modalities/checkpointing/torch/torch_checkpoint_loading.py
```diff
@@ -65,18 +65,15 @@ def load_model_checkpoint(self, model: nn.Module, file_path: Path) -> nn.Module:
         )
         return model
 
-    def load_optimizer_checkpoint(self, optimizer: Optimizer, model: nn.Module, file_path: Path) -> Optimizer:
+    def load_optimizer_checkpoint_(self, optimizer: Optimizer, model: nn.Module, file_path: Path):
         """
-        Load the optimizer checkpoint from the specified file path.
+        Load the optimizer checkpoint from the specified file path (in-plce).
```
mali-git, line 70:
```suggestion
        Load the optimizer checkpoint from the specified file path (in-place).
```

src/modalities/config/config.py
```diff
@@ -203,18 +223,23 @@ class CosineAnnealingLRSchedulerConfig(BaseModel):
...
-    checkpoint_loading: PydanticCheckpointLoadingIFType
+    checkpoint_loading: PydanticFSDP1CheckpointLoadingIFType
     checkpoint_path: Path
     wrapped_model: PydanticPytorchModuleType
     optimizer: PydanticOptimizerIFType
 
 
-class CheckpointedModelConfig(BaseModel):
-    checkpoint_loading: PydanticCheckpointLoadingIFType
+class FSDP1CheckpointedModelConfig(BaseModel):
```
mali-git, line 232: Here is a maming inconsitency between CheckpointedOptimizerConfig and FSDP1CheckpointedModelConfig. Please name either both with FSDP1 prefix or none with the prefix.

src/modalities/config/config.py
```diff
@@ -364,7 +428,9 @@ class RichResultSubscriberConfig(BaseModel):
 
 
 def load_app_config_dict(
-    config_file_path: Path, additional_resolver_funs: Optional[dict[str, Callable]] = None
+    config_file_path: Path,
+    experiment_id: Optional[str] = None,
```
mali-git, line 432: experimente_id in docstring missing.

src/modalities/config/config.py
```diff
@@ -382,11 +448,9 @@ def cuda_env_resolver_fun(var_name: str) -> int:
         int_env_variable_names = ["LOCAL_RANK", "WORLD_SIZE", "RANK"]
         return int(os.getenv(var_name)) if var_name in int_env_variable_names else os.getenv(var_name)
 
-    def modalities_env_resolver_fun(var_name: str, config_file_path: Path) -> str | Path:
-        if var_name == "experiment_id":
-            return get_experiment_id_of_run(config_file_path=config_file_path)
-        elif var_name == "config_file_path":
-            return config_file_path
+    def modalities_env_resolver_fun(var_name: str, kwargs: Path | str) -> str | Path:
```
mali-git, line 451: It seems that kwargs is a dictionary not a Path or string.

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -831,22 +854,28 @@ def __init__(
...
+                            # a meta device!
+                            attention_norm=attention_norm_config.norm_type.value(**dict(attention_norm_config.config)),
+                            ffn_norm=ffn_norm_config.norm_type.value(**dict(ffn_norm_config.config)),
                         )
                         for _ in range(n_layer)
                     ]
                 ),
-                lm_head_norm=lm_head_norm,
+                lm_head_norm=lm_head_norm_config.norm_type.value(**dict(lm_head_norm_config.config)),
+                lm_head=nn.Linear(in_features=n_embd, out_features=vocab_size, bias=False),
```
mali-git, line 867: Should bias not be configurable?

src/modalities/models/model_factory.py
```diff
@@ -1,33 +1,63 @@
...
 class ModelFactory:
     """Model factory class to create models."""
 
     @staticmethod
-    def get_checkpointed_model(
-        checkpoint_loading: CheckpointLoadingIF,
+    def _is_model_on_meta_device(model: nn.Module) -> bool:
+        meta_counter = 0
+        param_counter = 0
+        for name, tensor in itertools.chain(model.named_parameters(), model.named_buffers()):
```
mali-git, line 40:
```suggestion
        for _, tensor in itertools.chain(model.named_parameters(), model.named_buffers()):
```

src/modalities/models/model_factory.py
```diff
@@ -88,13 +126,69 @@ def get_fsdp_wrapped_model(
...
+
+        fsdp_config = {"mesh": device_mesh[ParallelismDegrees.DP_SHARD.value], "mp_policy": mp_policy}
+
+        modules = list(model.modules())
+        # we first shard all the blocks
+        for module_id, module in enumerate(modules):
+            if isinstance(module, block_types):
+                # As an optimization, we do not reshard after forward for the last
+                # transformer block since FSDP would prefetch it immediately.
+                reshard_block_after_forward = reshard_after_forward and int(module_id) < len(modules) - 1
```
mali-git, line 176: Why can we not reshard the last module?

src/modalities/models/model_factory.py
```diff
@@ -103,7 +197,26 @@ def get_weight_initalized_model(model: nn.Module, model_initializer: ModelInitia
...
+                module.reset_parameters()
+
+        # initialize the weights if they are on a meta device
+        if ModelFactory._is_model_on_meta_device(model=model):
+            # Allocate buffers and sharded parameters on GPU
+            model = model.to_empty(device="cuda")
+
+        # call reset_parameters on all nn.Modules that implement this function
+        # (normally all norms)
+        with torch.no_grad():
```
mali-git, line 217: Why do we need no_grad here?

src/modalities/models/model_factory.py
```diff
@@ -132,3 +245,83 @@ def get_activation_checkpointed_model(model: FSDP, activation_checkpointing_modu
...
+
+    @staticmethod
+    def get_compiled_model(model: nn.Module, block_names: list[str]) -> nn.Module:
+        """
+        Apply torch.compile to each transformer block, which makes compilation efficient due to
+        repeated structure. Alternatively one can compile the whole model (after applying DP).
+        Inspired by: https://github.com/pytorch/torchtitan/blob/6b2912a9b53464bfef744e62100716271b2b248f/torchtitan/parallelisms/parallelize_llama.py#L275
+        """
+
+        def get_parent_module_and_child_name(child_module: nn.Module, model: nn.Module) -> tuple[nn.Module, str]:
```
mali-git, line 257: It does not return child_name. Please rename function.

src/modalities/models/model_factory.py
```diff
@@ -132,3 +245,83 @@ def get_activation_checkpointed_model(model: FSDP, activation_checkpointing_modu
...
+                for child_name, child_candidate in parent_candidate.named_children():
+                    if child_candidate is child_module:
+                        return parent_candidate, child_name
+            raise ModelStateError("No valid parent candidate")
+
+        block_types = tuple([get_module_class_from_name(model, b) for b in block_names])
+
+        for _, module in model.named_modules():
+            if isinstance(module, block_types):
+                compiled_module = torch.compile(module, fullgraph=True)
```
mali-git, line 268: Please add a comment explaining full_graph=True.

src/modalities/models/model_factory.py
```diff
@@ -132,3 +245,83 @@ def get_activation_checkpointed_model(model: FSDP, activation_checkpointing_modu
...
+                parent_module, child_name = get_parent_module_and_child_name(child_module=module, model=model)
+                parent_module.register_module(name=child_name, module=compiled_module)
```
mali-git, line 270: Why do we need to register the module?

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -39,13 +39,16 @@ def get_adam_w(
...
-    def get_checkpointed_optimizer(
-        checkpoint_loading: CheckpointLoadingIF, checkpoint_path: Path, wrapped_model: nn.Module, optimizer: Optimizer
+    def get_checkpointed_optimizer_(
+        checkpoint_loading: FSDP1CheckpointLoadingIF,
+        checkpoint_path: Path,
+        wrapped_model: nn.Module,
+        optimizer: Optimizer,
     ) -> Optimizer:
-        wrapped_optimizer = checkpoint_loading.load_optimizer_checkpoint(
+        checkpoint_loading.load_optimizer_checkpoint_(
```
mali-git, line 48: Is the checkpoint loaded in-place?

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -84,7 +89,12 @@ def _assert_existence_of_weight_decay_groups_excluded(model: FSDP, weight_decay_
         }
         weight_decay_groups_excluded = ["embedding", "layernorm"]
     """
-    nn_model: NNModel = model.module
+    # FSDP 1
```
mali-git, line 92: We could encapsulate the logic in a helper function since it is used at least twice.

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -79,6 +88,130 @@ def clip_gradients(self) -> torch.Tensor:
...
+        # If total_norm is a DTensor, the placements must be `torch.distributed._tensor.ops.math_ops._NormPartial`.
+        # We can simply reduce the DTensor to get the total norm in this tensor's process group
+        # and then convert it to a local tensor.
+        # NOTE: It has the purpose to return a reduced total_norm tensor whose .item() would return the correct value
+        if isinstance(total_norm, DTensor):
+            # Will reach here if any non-PP parallelism is used.
+            # If only using PP, total_norm will be a local tensor.
+            total_norm = total_norm.full_tensor()
+
+        torch.nn.utils.clip_grads_with_norm_(parameters, max_norm, total_norm, foreach)
```
mali-git, line 174: Please use named arguments.

src/modalities/utils/mfu.py
```diff
@@ -37,53 +39,59 @@ def _get_theoretical_gpu_peak_performance_single(precision: torch.dtype, gpu_typ
...
-        if model.mixed_precision.reduce_dtype != precision or model.mixed_precision.buffer_dtype != precision:
-            warnings.warn(f"Could not get theoretical GPU peak performance for mixed precision type = {precision}.")
+        if isinstance(model, FSDP1):
+            precision = model.mixed_precision.param_dtype
+            if model.mixed_precision.reduce_dtype != precision or model.mixed_precision.buffer_dtype != precision:
+                warnings.warn(f"Could not get theoretical GPU peak performance for mixed precision type = {precision}.")
+                return None
+        elif isinstance(model, FSDP2):
+            warnings.warn("MFU is computed based on the assumption that bf16 precision is used.")
+            precision = torch.bfloat16
```
mali-git, line 63: We also plan to use FP8.
TODO: Create an issue.

src/modalities/utils/mfu.py
```diff
@@ -37,53 +39,59 @@ def _get_theoretical_gpu_peak_performance_single(precision: torch.dtype, gpu_typ
...
+        else:
+            raise TypeError(f"Model should be of type FSDPX, but is {type(model)} instead.")
+
+        device_name = torch.cuda.get_device_name()
+        if device_name.startswith("NVIDIA A100"):
+            single_gpu_peak_performance = _get_theoretical_gpu_peak_performance_single(precision, "A100")
+        elif device_name.startswith("NVIDIA H100"):
+            single_gpu_peak_performance = _get_theoretical_gpu_peak_performance_single(precision, "H100")
+        else:
```
mali-git, line 72: TODO: Create an issue for GH200 support.

src/modalities/trainer.py
```diff
@@ -279,8 +279,8 @@ def train(
                     # TODO: hardcoded metric key
                     throughput_metrics={
                         "train samples/s": ResultItem(synced_num_samples_per_second, 1),
-                        "train mfu": ResultItem(mfu, 2),
-                        "lr mean": ResultItem(torch.tensor(scheduler.get_last_lr()).mean()),
+                        "train mfu (16-bit)": ResultItem(mfu, 2),
```
mali-git, line 282: TODO: Support FP8

src/modalities/util.py
```diff
@@ -43,29 +45,60 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
...
-        num_local_train_micro_batches_ret % gradient_acc_steps == 0
-    ), "callback_interval_in_batches_per_rank must be divisible by gradient_acc_steps"
-    return num_local_train_micro_batches_ret
+def get_experiment_id_of_run(
+    config_file_path: Path, hash_length: Optional[int] = 8, max_experment_id_byte_length: Optional[int] = 1024
+) -> str:
+    """Create a unique experiment ID for the current run on rank 0 and broadcast it to all ranks.
+    Internally, the experiment ID is generated by hashing the configuration file path and appending
+    the current date and time.
+    The experiment ID is then converted to a byte array (with maximum length of max_experment_id_byte_length) and
```
mali-git, line 54:
```suggestion
    The experiment ID is then converted to a byte array (with maximum length of max_experiment_id_byte_length) and
```

src/modalities/util.py
```diff
@@ -43,29 +45,60 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
...
+    broadcasted to all ranks. In the unlikely case of the experiment ID being too long, a ValueError is raised
+    and max_experment_id_byte_length must be increased. Each rank then decodes the byte array to the original
+    string representation and returns it. Having a globally synced experiment ID is mandatory for
+    saving files / checkpionts in a distributed training setup.
 
+    Args:
+        config_file_path (Path): Path to the configuration file.
+        hash_length (Optional[int], optional): Defines the char length of the commit hash. Defaults to 8.
+        max_experment_id_byte_length (Optional[int], optional): Defines max byte length of the experiment_id
```
mali-git, line 63:
```suggestion
        max_experiment_id_byte_length (Optional[int], optional): Defines max byte length of the experiment_id
```

mali-git (changes requested):
Great work!

Only small adjustments are needed.

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -20,15 +24,18 @@ class GradientClippingMode(LookupEnum):
     MAX_NORM = "inf"
```
mali-git, line 24: Is `float("inf")` safer?

src/modalities/models/gpt2/gpt2_model.py
le1nux: I would leave it as is for now. Before, we were tying lm_head with the embedding weights. In this case, there was also no bias term.

src/modalities/models/model_factory.py
le1nux: This seems already correct to me. Was this corrected maybe already in the mean-time?

src/modalities/optimizers/optimizer_factory.py
le1nux: We need to return this, since it is used as a component. 
https://github.com/Modalities/modalities/blob/ef2f8c27ae6b36c65e5cf3600127e3cc3f0b033e/src/modalities/registry/components.py#L172

src/modalities/utils/mfu.py
le1nux: done

src/modalities/models/model_factory.py
le1nux: To maximize speedup this is recommended. 
https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html


> In order to maximize speedup, graph breaks should be limited. We can force TorchDynamo to raise an error upon the first graph break encountered by using 
```
fullgraph=True:

opt_bar = torch.compile(bar, fullgraph=True)
try:
    opt_bar(torch.randn(10), torch.randn(10))
except:
    tb.print_exc()
```

src/modalities/trainer.py
le1nux: I created an issue for this.

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,197 @@
...
+        set_model_state_dict(model=app_state.model, model_state_dict=state_dict, options=StateDictOptions(strict=False))
+
+
+class OptimizerStateRetriever(StateRetrieverIF):
+    @staticmethod
+    def get_state_dict(app_state: AppState) -> dict[str, Any]:
+        sd = get_optimizer_state_dict(
+            model=app_state.model,
+            optimizers=app_state.optimizer,
+            options=StateDictOptions(flatten_optimizer_state_dict=True),
```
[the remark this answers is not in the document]
le1nux: Torch Titan had documentation on this in their code base:
https://github.com/pytorch/torchtitan/blob/b291ad662493b63d25b038a30a915082d3617baf/torchtitan/components/checkpoint.py#L193-L214

Since we want to have PP support soon, we also have to flatten the optimizer state dict.

src/modalities/checkpointing/checkpoint_loading.py
le1nux: With the current setup, FSDP2 always requires DCP and with FSDP1 we resort to the old checkpointing implementation (as before). 
Since we will throw out FSDP1 eventually, I renamed this IF to be FSDP1-specific (so that FSDP2 is never used with this checkpointing implementation).

docs/dev_ops/tests.md
```diff
@@ -0,0 +1,25 @@
+# Testing Modalities
+
+Modalities has a threefold setup for testing, namely
+
+* Main tests <br> 
+The main tests comprise CPU, single GPU and multi-GPU tests. The latter ones create a distributed environment internally and allow end2end testing of modalities. 
+Each of these tests defines its requirements (typically the number of GPUs) in the test and the test will be skipped if the requirements are not met.
+
+* Torchrun tests <br>
+These tests are run from a shell script using torchrun and are typically end2end or at least integration tests. Since we implemented distributed testing using multiprocessing within modalities, these tests will be integrated into the main tests in the long term. Note that some of the torchrun tests have been already migrated to the main tests.
```
mali-git, line 10:
```suggestion
These tests are run from a shell script using torchrun and are typically end2end or at least integration tests. Since we implemented distributed testing using multiprocessing within Modalities, these tests will be integrated into the main tests in the long term. Note that some of the torchrun tests have been already migrated to the main tests. 
```

docs/fsdp1_vs_fsdp_2.md
```diff
@@ -0,0 +1,115 @@
...
+The main difference in the instantiation dependencies between FSDP1 and FSDP2 is that in FSDP1, the model is wrapped after the weights are initialized, while in FSDP2, the model is wrapped before the weights are initialized. Therefore, FSDP1 requires the model to be fully materialized in CPU RAM before wrapping, as otherwise the weights cannot be initialized. In the case of FSDP2, we can have the model on a meta device, shard it and then move it to the data device during the initialization. This way, only the shards are materialized in GPU RAM and we have an extremely low CPU RAM footprint. 
+See also the torch titan [documentation](https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md) on the differences between FSDP1 and FSDP2. 
+
+
+# Instantiation dependencies (Warmstart)
+
+## FSDP1
+Similar to before, we first instantiate the model (random weights), initialize the weights following a pre-defined distribution, and load the actual weights from the checkpoint and shard the model with FSDP1. Loading and sharding is one step in modalities. We first load the model into the CPU RAM on rank 0 and then wrap the model with FSDP1, sharding the model to the other ranks.
```
mali-git, line 57:
```suggestion
Similar to before, we first instantiate the model (random weights), initialize the weights following a pre-defined distribution, and load the actual weights from the checkpoint and shard the model with FSDP1. Loading and sharding is one step in Modalities. We first load the model into the CPU RAM on rank 0 and then wrap the model with FSDP1, sharding the model to the other ranks.
```

src/modalities/checkpointing/stateful/app_state.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @staticmethod
+    def load_state_dict_(app_state: AppState, state_dict: dict[str, Any]) -> None:
+        """Loads the state dict into the model in the AppState object.
+
+        Args:
+            app_state (AppState): The app_state object containing the model.
+            state_dict (dict[str, Any]): The state dict to load into the model.
+        """
+        set_model_state_dict(model=app_state.model, model_state_dict=state_dict, options=StateDictOptions(strict=False))
```
mali-git, line 174: Why is strict per default equal to False? Please add a comment.

src/modalities/checkpointing/fsdp/fsdp_checkpoint_loading.py
```diff
@@ -38,6 +40,7 @@ def __init__(
         self.mixed_precision_settings = mixed_precision_settings
         self.sharding_strategy = sharding_strategy
 
+    @torch.no_grad()
```
mali-git, line 43:
```suggestion
    # Ensures reduced memory footprint and avoids side-effects
    @torch.no_grad()
```

src/modalities/config/component_factory.py
```diff
@@ -30,13 +30,32 @@ def build_components(self, config_dict: dict, components_model_type: Type[BaseMo
         Returns:
             BaseModelChild: Instance of the components_model_type with the built components.
         """
-        component_names = list(components_model_type.model_fields.keys())
-        component_dict = self._build_config(config_dict=config_dict, component_names=component_names)
+        # the componentsinstantiaton model allows for the definition of required and optional top-level components
```
mali-git, line 33:
```suggestion
        # the components instantiaton model allows for the definition of required and optional top-level components
```

src/modalities/config/component_factory.py
```diff
@@ -30,13 +30,32 @@ def build_components(self, config_dict: dict, components_model_type: Type[BaseMo
...
+        # for example the mfu_calculator might not always be required.
+        component_names_required = [
```
mali-git, line 35: component_names = dict(required=[], optional=[])

for name, field in components_model_type.model_fields.items():
    (component_names["required"] if field.is_required() else component_names["optional"]).append(name)

src/modalities/models/components/layer_norms.py
```diff
@@ -57,18 +55,10 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-    Enum lookup class for LayerNorms.
-
-    Attributes:
-        RMSNorm: RMSLayerNorm class.
-        LayerNorm: nn.LayerNorm class.
-    """
-
-    RMSNorm = RMSLayerNorm
-    LayerNorm = nn.LayerNorm
+    def reset_parameters(self):
```
mali-git, line 58: Can we add more information describing why this is necessary?

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -831,22 +854,30 @@ def __init__(
                             attention_config=attention_config,
                             dropout=dropout,
                             ffn_hidden=ffn_hidden,
-                            attention_norm=deepcopy(attention_norm),
-                            ffn_norm=deepcopy(ffn_norm),
+                            # deepcopy did not work here! The weights were then automatically
+                            # moved to a cuda device even when the deepcopied weights were on
+                            # a meta device!
+                            attention_norm=attention_norm_config.norm_type.value(**dict(attention_norm_config.config)),
```
mali-git, line 860: Here, we could use a factory.

Outside of GPT2LLM:

def make_norm(norm_config):
    norm_cls = norm_config.norm_type.value
    return lambda: norm_cls(**norm_config.config)

attention_norm_factory = make_norm(attention_norm_config)
ffn_norm_factory = make_norm(ffn_norm_config)

Inside GPT2LLM:

h=nn.ModuleList(
    [
        GPT2Block(
            ...,
            attention_norm=attention_norm_factory(),
            ffn_norm=ffn_norm_factory(),
        )
        for _ in range(n_layer)
    ]
)

src/modalities/models/model_factory.py
```diff
@@ -103,7 +202,26 @@ def get_weight_initalized_model(model: nn.Module, model_initializer: ModelInitia
...
+            for submodule in module.children():
+                reset_parameters_if_function_exists(submodule)
+
+            # Check if the module has the `reset_parameters` method
+            if hasattr(module, "reset_parameters") and callable(getattr(module, "reset_parameters")):
+                module.reset_parameters()
+
+        if ModelFactory._is_model_on_meta_device(model=model):
+            # Allocate buffers and sharded parameters on GPU
+            model = model.to_empty(device="cuda")
```
mali-git, line 217: Is my understanding correct that the model is a FSDP2 wrapped model since it is assumed that the model is already sharded?

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -95,20 +109,42 @@ def _assert_existence_of_weight_decay_groups_excluded(model: FSDP, weight_decay_
...
-    model: FSDP, weight_decay: float, weight_decay_groups_excluded: list[str]
+    model: FSDPX, weight_decay: float, weight_decay_groups_excluded: list[str]
 ) -> tuple[OptimizerGroups, list[str]]:
     """
     create optimizer groups of parameters with different weight decays that are to be used in Adam or AdamW
     """
-    nn_model: NNModel = model.module
-    weight_decay_groups = nn_model.weight_decay_groups
-    params = {name: parameter for name, parameter in model.named_parameters() if parameter.requires_grad}
+    # FSDP 1
```
mali-git, line 117: Re-use logic:

if isinstance(model, (FSDP1, FSDP2)):
    nn_model = model.module if isinstance(model, FSDP1) else model
    weight_decay_groups = nn_model.weight_decay_groups

    def is_valid_param(name, param):
        if not param.requires_grad:
            return False
        if isinstance(model, FSDP2) and isinstance(param, DTensor):
            return param.to_local().numel() > 0
        return True

    params = {
        name: param
        for name, param in model.named_parameters()
        if is_valid_param(name, param)
    }

else:
    raise OptimizerError(
        f"model {type(model)} is not an instance of FSDP1 or FSDP2. Please use the correct model type."
    )

src/modalities/running_env/fsdp/device_mesh.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @model_validator(mode="after")
+    def _validate(self):
+        if (
+            self.tensor_parallel_degree != 1
+            or self.pipeline_parallel_degree != 1
+            or self.context_parallel_degree != 1
+            or self.enable_loss_parallel
+        ):
+            raise ConfigError("Only tensor parallelism, data parallelism and loss parallelism are not supported, yet.")
```
mali-git, line 31: "Tensor parallelism, context parallelism and loss parallelism are not supported, yet." ?

src/modalities/util.py
```diff
@@ -43,29 +47,60 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
...
+                f"Experiment ID is too long: {len(experimenet_id_bytes)} bytes, "
+                f"max length is {max_experment_id_byte_length} bytes"
+            )
+        print(f"Rank 0 generated experiment_id: {experimenet_id}")
+    else:
+        experimenet_id_bytes = bytearray(max_experment_id_byte_length)  # Preallocate buffer for receiving
+
+    # Ensure all ranks have the same folder name
+    experment_id_tensor = torch.tensor(
+        list(experimenet_id_bytes) + [0] * (max_experment_id_byte_length - len(experimenet_id_bytes)), dtype=torch.uint8
```
mali-git, line 97:
```suggestion
        list(experimenet_id_bytes) + [0] * (max_experiment_id_byte_length - len(experimenet_id_bytes)), dtype=torch.uint8
```

src/modalities/util.py
```diff
@@ -43,29 +47,60 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
...
+    ).cuda()
+    dist.broadcast(experment_id_tensor, src=0)
+
+    # Decode on all ranks
+    experiment_id = experment_id_tensor.cpu().numpy().tobytes().decode("utf-8").rstrip("\x00")
```
mali-git, line 102: If experiment_id_bytes endes with zero bytes (\x00, we strip parts of the id.
Broadcast also the lenght of the id?

src/modalities/util.py
```diff
@@ -77,26 +112,64 @@ def format_metrics_to_gb(item):
...
+        if model.sharding_strategy.name == "NO_SHARD":
+            sharding_factor = dist.get_world_size()
+        if model.sharding_strategy.name == "HYBRID_SHARD":
+            sharding_factor = dist.get_world_size() // torch.cuda.device_count()
+        elif model.sharding_strategy.name == "FULL_SHARD":
+            sharding_factor = 1
+        total_num_params = total_num_params // sharding_factor
+        return total_num_params
+    elif isinstance(model, FSDP2):
+        total_num_params = sum(p.numel() for p in model.parameters() if p.requires_grad)
```
mali-git, line 166: Why do we not need an all_reduce of the local parameters?

src/modalities/utils/mfu.py
```diff
@@ -18,124 +20,171 @@
...
+        """
+        Computes the MFU for the given number of samples per second.
+
+        Args:
+            num_samples_per_second (torch.Tensor): The number of samples per second.
+
+        Returns:
+            torch.Tensor: The computed MFU.
+        """
+        return MFUCalculatorABC._compute_mfu_impl(
```
mali-git, line 185: Please use named arguments.

src/modalities/checkpointing/stateful/app_state.py
le1nux: TorchTitan also sets this to False: https://github.com/pytorch/torchtitan/blob/30b9ea0b1ad893379d2ff3b12dbf18600730c249/torchtitan/components/checkpoint.py#L63

However, I'm not entirely sure why. 

There is also not really any documentation on it:
https://docs.pytorch.org/docs/stable/distributed.checkpoint.html#torch.distributed.checkpoint.state_dict.StateDictOptions

src/modalities/models/gpt2/gpt2_model.py
le1nux: Should we integrate that at a later step, maybe? 
I tried to make a minimal change w.r.t. norm instantiation.

src/modalities/models/model_factory.py
le1nux: yes, see our documentation here:
[In the case of FSDP2, we first instantiate the model, wrap it with FSDP2 and then initialize the weights following a pre-defined distribution. The model_raw can be instantiated on a meta device. After sharding with FSDP2 (wrapped_model), the model is still on a meta device and only during the initialisation, the model is moved to the data device (see [here](https://github.com/Modalities/modalities/blob/a7f683a6202209dcf8e78cbff4ca68991e89e081/src/modalities/models/model_factory.py)).](https://github.com/Modalities/modalities/blob/b915fd1e6f2bdb5044b5d71dd455b5518accaf21/docs/fsdp1_vs_fsdp_2.md#L27-L28)

src/modalities/optimizers/optimizer_factory.py
le1nux: In my opinion the code duplication is very limited and with the introduction of the checks within `is_valid_params` it makes it rather difficult to read. I think having the if condition based on FSDP1 and FSDP2 is fine, since we have different filtering routines.

src/modalities/running_env/fsdp/device_mesh.py
le1nux: We only allow a two dimensional device mesh for now, i.e. replication and data parallel. The other dimensions are needed for future parallelism strategies such as TP or PP. I added them already to the IF though, since we will integrate in the next days.

src/modalities/util.py
le1nux: If a string "0" is encoded via utf-8, it has the byte representation \x30. 
\x00 is a non-printable control character, which can be used to indicate the end of a sequence. 
https://www.utf8-chartable.de/
https://en.wikipedia.org/wiki/Null_character

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -39,13 +43,27 @@ def get_adam_w(
         return optimizer
 
     @staticmethod
-    def get_checkpointed_optimizer(
-        checkpoint_loading: CheckpointLoadingIF, checkpoint_path: Path, wrapped_model: nn.Module, optimizer: Optimizer
+    def get_fsdp1_checkpointed_optimizer_(
```
flxst, line 46: What is the purpose of the trailing underscore here? 

If I understand correctly, this is generally used in our codebase in order to indicate that an argument of a method is modified in-place (see e.g. line 63). However, the present method returns the `optimizer` explicitly.

src/modalities/running_env/fsdp/device_mesh.py
flxst: My interpretation of @mali-git's comment is that the word "Only" is redundant / confusing. We might also want to add PP to the sentence: "Tensor parallelism, pipeline parallelism, context parallelism and loss parallelism are not supported yet."

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -20,15 +24,18 @@ class GradientClippingMode(LookupEnum):
...
 
 
-class FSDPGradientClipper(GradientClipperIF):
-    """The FSDPGradientClipper class that is responsible for clipping the gradients of a model wrapped with FSDP."""
+class FSDP1GradientClipper(GradientClipperIF):
+    """The FSDP1GradientClipper class that is responsible for clipping the gradients of a model wrapped with FSDP.
+    Follows the documeentation from
```
flxst, line 29:
```suggestion
    Follows the documentation from
```

src/modalities/models/model_factory.py
```diff
@@ -1,33 +1,66 @@
...
+        if meta_counter > 0 and meta_counter < param_counter:
+            raise ModelStateError("Either all or none of the parameters and buffers must be on meta device!")
+        return meta_counter > 0
+
+    @staticmethod
+    def get_fsdp1_checkpointed_model(
+        checkpoint_loading: FSDP1CheckpointLoadingIF,
         checkpoint_path: Path,
         model: nn.Module,
     ) -> nn.Module:
```
flxst, line 57:
```suggestion
    ) -> FSDP:
```

src/modalities/models/components/layer_norms.py
```diff
@@ -57,18 +55,14 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        # This function is called from the ModelFactory, when the model weights are initialized.
+        # We recursively iterate through all the modules of a model and call reset_parameters if it exists.
+        # Subsequently, we apply run the weight initialization on top, applying custom weight initialization to
+        # a subset of modules based on the configuration.
+        # Inpired by torch titan RMS Norm implementation:
+        # https://github.com/pytorch/torchtitan/blob/de9fd2b9ea7e763c9182e0df81fc32c2618cc0b6/torchtitan/models/norms.py#L113C1-L114C57
+        torch.nn.init.ones_(self.weight)
```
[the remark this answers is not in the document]
le1nux: fixed in 5283321fec91718e371ed94af261aa87566bf84f

src/modalities/optimizers/optimizer_factory.py
le1nux: That is correct. We load the checkpoint in-place (as indicated by the trailing _) but return the modified optimizer nevertheless. We need to do this, to pass the optimizer to another component, as defined in the dependency graph in the yaml config.

flxst (approved):
LGTM! :)

src/modalities/models/gpt2/gpt2_model.py
mali-git: Sounds good - Please create an issue.

src/modalities/running_env/fsdp/device_mesh.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @model_validator(mode="after")
+    def _validate(self):
+        if (
+            self.tensor_parallel_degree != 1
+            or self.pipeline_parallel_degree != 1
+            or self.context_parallel_degree != 1
+            or self.enable_loss_parallel
+        ):
+            logger.warning(
```
mali-git, line 35: Why not raising an exception?

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -79,6 +88,135 @@ def clip_gradients(self) -> torch.Tensor:
...
+        total_norm = torch.nn.utils.get_total_norm(
+            tensors=grads, norm_type=norm_type, error_if_nonfinite=error_if_nonfinite, foreach=foreach
+        )
+
+        # Inspired by torch titan
+        # If total_norm is a DTensor, the placements must be `torch.distributed._tensor.ops.math_ops._NormPartial`.
+        # We can simply reduce the DTensor to get the total norm in this tensor's process group
+        # and then convert it to a local tensor.
+        # NOTE: It has the purpose to return a reduced total_norm tensor whose .item() would return the correct value
+        if isinstance(total_norm, DTensor):
```
mali-git, line 172: To my understanding, the current approach does not work for PP. If so, it would be good to raise an Exception when PP > 1.

src/modalities/utils/mfu.py
```diff
@@ -18,124 +20,171 @@
...
+                or wrapped_model.mixed_precision.buffer_dtype != precision
+            ):
+                warnings.warn(f"Could not get theoretical GPU peak performance for mixed precision type = {precision}.")
                 return None
-            else:
-                return single_gpu_peak_performance * world_size
-    else:
-        return None
+        elif isinstance(wrapped_model, FSDP2):
+            warnings.warn("MFU is computed based on the assumption that bf16 precision is used.")
```
mali-git, line 117: Add a comment describing that dtype cannot be inferred from wrapped_model in FSDP2, and therefore, we assume that BF16 is used?

src/modalities/util.py
```diff
@@ -43,29 +47,60 @@ def parse_enum_by_name(name: str, enum_type: Type[Enum]) -> Enum:
...
+    if rank == 0:
+        # Generate a unique folder name
+        experimenet_id = get_experiment_id_from_config(config_file_path, hash_length)
+        experiment_id_bytes = experimenet_id.encode("utf-8")
+        if len(experiment_id_bytes) > max_experiment_id_byte_length:
+            raise ValueError(
+                f"Experiment ID is too long: {len(experiment_id_bytes)} bytes, "
+                f"max length is {max_experiment_id_byte_length} bytes"
+            )
+        print(f"Rank 0 generated experiment_id: {experimenet_id}")
```
mali-git, line 91: Use logging instead of print.

mali-git (approved):

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
le1nux: Torch titan passes in the pp_mesh. We don't have this anyways yet and throw an exception/raise a warning if someone sets pp > 1 in the device mesh

src/modalities/trainer.py
```diff
@@ -266,21 +269,17 @@ def train(
...
+                    mfu_score = self.mfu_calculator.compute(num_samples_per_second=synced_num_samples_per_second)
                 training_metrics = EvaluationResultBatch(
                     losses=losses,
                     metrics=metrics,
                     # TODO: hardcoded metric key
                     throughput_metrics={
                         "train samples/s": ResultItem(synced_num_samples_per_second, 1),
-                        "train mfu": ResultItem(mfu, 2),
-                        "lr mean": ResultItem(torch.tensor(scheduler.get_last_lr()).mean()),
+                        "train mfu (16-bit)": ResultItem(mfu_score, 2),
```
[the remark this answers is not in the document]
le1nux: yes, we can only calculate MFU w.r.t. to 16 bit. @flxst investigated this.

src/modalities/running_env/fsdp/device_mesh.py
le1nux: I  made it a warning so we could properly test the device mesh. I will make it configurable to raise or warn.

PR #317, opened 2025-03-11
FSDP2: Fix MFU

flxst:
# What does this PR do?

This PR
- fixes the MFU computation for FSDP2 (based on the assumption that BF16 is used)
- updates training config files to account for changes in `attention_norm_config` & `checkpoint_saving_execution`
- sketches potential housekeeping / type annotations with FSDP1 & FSDP2 (will be continued in follow-up work)

Note: The MFU Test has not completely been fixed. This will also be done in follow-up work.

## General Changes
* `config_lorem_ipsum.yaml` now uses LayerNorm instead of RMSNorm

## Breaking Changes
* None


src/modalities/trainer.py
```diff
@@ -279,7 +279,7 @@ def train(
                     # TODO: hardcoded metric key
                     throughput_metrics={
                         "train samples/s": ResultItem(synced_num_samples_per_second, 1),
-                        "train mfu": ResultItem(mfu, 2),
+                        "train mfu (16-bit)": ResultItem(mfu, 2),
```
le1nux, line 282: would there be a difference in bf16 to fp16? If so, I would rename this to BF16 to be more precise.

src/modalities/utils/mfu.py
```diff
@@ -38,47 +42,47 @@ def _get_theoretical_gpu_peak_performance_single(precision: torch.dtype, gpu_typ
...
-    #             return single_gpu_peak_performance * world_size
-    # else:
-    #     return None
-
-    return None
+    if torch.cuda.is_available() and torch.cuda.device_count() > 0:
+        if isinstance(model, FSDP1):
+            precision = model.mixed_precision.param_dtype
+            if model.mixed_precision.reduce_dtype != precision or model.mixed_precision.buffer_dtype != precision:
+                warnings.warn(f"Could not get theoretical GPU peak performance for mixed precision type = {precision}.")
```
le1nux, line 60: Isn't this warning misleading? 
From the if statement I would have expected a warning such as:
"Could not get theoretical GPU peak performance. In mixed precision, the param_dtype must be equal to reduce_dtype and buffer_dtype."

le1nux (approved):
LGTM! :) 
Just left a minor comment.

src/modalities/trainer.py
flxst: No, the computation is the same for fp16 and bf16, that's why I chose the general name 16-bit.

PR #346, opened 2025-04-11
MFU calculation  support with FSDP2

le1nux:
# What does this PR do?

This PR adds support for MFU calculation with FSDP2

## General Changes
* ..

## Breaking Changes
* .. 

### DIFF
--- a/src/modalities/optimizers/optimizer_factory.py
+++ b/src/modalities/optimizers/optimizer_factory.py
@@ -2,13 +2,17 @@ import re
 from pathlib import Path
 
 import torch.nn as nn
+from torch.distributed.fsdp import FSDPModule as FSDP2
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP1
+from torch.distributed.tensor import DTensor
 from torch.optim import Adam, AdamW, Optimizer
 
 from modalities.checkpointing.checkpoint_loading import FSDP1CheckpointLoadingIF
 from modalities.exceptions import OptimizerError
 from modalities.models.model import NNModel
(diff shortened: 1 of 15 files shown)


src/modalities/optimizers/optimizer_factory.py
```diff
@@ -105,19 +109,31 @@ def _assert_existence_of_weight_decay_groups_excluded(
...
 ) -> tuple[OptimizerGroups, list[str]]:
     """
     create optimizer groups of parameters with different weight decays that are to be used in Adam or AdamW
     """
     # FSDP 1
-    if hasattr(model, "module"):
+    if isinstance(model, FSDP1):
         nn_model: NNModel = model.module
+        weight_decay_groups = nn_model.weight_decay_groups
+        params = {name: parameter for name, parameter in model.named_parameters() if parameter.requires_grad}
```
fromm-m, line 121: We could raise a warning if params are empty! Maybe also an error (could be problematic for inference?)

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -105,19 +109,31 @@ def _assert_existence_of_weight_decay_groups_excluded(
...
+
     # FSDP 2
-    else:
+    elif isinstance(model, FSDP2):
         nn_model = model
-    weight_decay_groups = nn_model.weight_decay_groups
-    params = {name: parameter for name, parameter in model.named_parameters() if parameter.requires_grad}
+        weight_decay_groups = nn_model.weight_decay_groups
+        params = {
```
fromm-m, line 127: Similarly, we could raise a warning if params are empty! Maybe also an error (could be problematic for inference?)

le1nux: Raising an error now 👍

PR #380, opened 2025-07-03
Instruction Tuning Improvements

le1nux:
# What does this PR do?

This PR ..

## General Changes
* ..

## Breaking Changes
* .. 

### DIFF
--- a/src/modalities/dataloader/collate_fns/collator.py
+++ b/src/modalities/dataloader/collate_fns/collator.py
@@ -70,11 +70,14 @@ class DefaultWrappingCollator(CollatorIF):
         Returns:
             DatasetBatch: The processed batch of data.
         """
+        # convert to tensors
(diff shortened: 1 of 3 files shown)


le1nux (commented):
This is a draft. 

What has been done so far: 
* Refactored the collator design which is now composable. Meaning you can call multiple collate_fns sequentially. This is helpful for instance for instruction tuning, where you first shift the targets for the autoregressive objective and then mask certiain tokens to disregard in the loss such as input from the user role. In this case, we would have two collate_fns being called one after the other. 

* We can specify now, which component to load in the app_state from a DCP checkpoint. This is important for continued pretraining or finetuning, where we don't want to use the previous optimizer and lr scheduler states. 

* Added an iterative mem map dataset implemenation. Instead of packing the samples, we take the index (stored in the pbin) to iterate over each sample. 

Todos:
* Add tests for the new functionality 
* Fix failing tests. 
* Verify correctness of the code.

PR #390, opened 2025-08-01
Fix of SwiGLU hidden not being multiple of world size

le1nux:
# What does this PR do?

This PR fixe an issue that occurs when using SwiGLU in a FSDP2 + TP combination. 
Since the hidden dim is computed dynamically to compensate for the difference in the number of trainable weights in comparison to GeLU, we end up with hidden dims that are not a multiple of the world size. 

This in return creates a problem when applying TP and subsequently FSDP2, as uneven sharding is not allowed for FSDP2 + TP (yet). The issue does not occur when only using FSDP2, as FSDP2 already support uneven sharding out of the box. 

## General Changes
* ..

## Breaking Changes
* .. 


fromm-m (approved):
LGTM

PR #396, opened 2025-08-25
Fixes eager conversion on GPU

CYHSM:
Reopening from https://github.com/Modalities/modalities/pull/392 now into main branch.  See original below:

# What does this PR do?

This PR fixes the conversion error we get when running the conversion in eager mode on GPU. It seems the upcasting to float32 with subsequent downcasting introduces errors into the attention_weights and therefore also the logits (see below plot). So it seems that historically this was added when using half-precision (fp16) as it has limited range and can easily overflow/underflow, which should be fixed with the extended range of BF16.

comparison_differences

The logits still are similar enough to produce similar outputs so thats why we see similar outputs even though the logits of the converted model are not the same
comparison_logits

Not confirmed on main due to import errors from the updated transformers library, so comparing to the updating_and_fixing_conversion branch instead. 

## General Changes
* Changes the calculation of attn_weights to not upcast to float 32

### DIFF
--- a/src/modalities/conversion/gpt2/modeling_gpt2.py
+++ b/src/modalities/conversion/gpt2/modeling_gpt2.py
@@ -161,7 +161,9 @@ def eager_attention_forward(
         causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
         attn_weights = attn_weights + causal_mask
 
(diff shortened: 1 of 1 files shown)


le1nux (approved):
LGTM :)

PR #399, opened 2025-09-17
Pipeline parallelism continued

BlueCrescent:
# What does this PR do?

- Issue: Dicts, as used in modalities as model/loss inputs and outputs, are not supported in the torch pp code.
  - Reason: Direct GPU to GPU communication
  - See: https://github.com/pytorch/pytorch/issues/159711
  - Workaround: Overloaded corresponding forward() and __call__() methods to allow direct tensor input and output.
- Issue: Loss accumulation (for logging) needs to use DP world size in stead of total world size (also TP and CP?)
  - Solution:
    - For the averaged loss we only increment the batch count on the last stages of a PP schedule.
    - For the current batch loss we added the dp world size, computed from the device mesh, to the Trainer class.
- Issue: Due to a Bug in PyTorch we cannot evaluate before training begin.
  - Workaround: Disabled eval before training
  - Solution: Issue and PR opened for PyTorch 2.9
  - See: https://github.com/pytorch/pytorch/issues/162822
- Issue: Normal distributed dataloader/sampler use would loose batches
  - Solution: Use (Resumable)DistributedMultiDimSamplerConfig
  - TODO: Implement DistributedMultiDimSamplerConfig for eval dataloader
  - TODO??: Remove data_parallel_key param from SamplerFactory.create_resumable_distributed_multi_dim_sampler()
    - Instead use BOTH ParallelismDegrees.DP_REPLICATE and ParallelismDegrees.DP_SHARD (multiplied)
- Issue: Gradient clipping needs to sync over all stages
  - Solution: Implemented same procedure using all_reduce() as in TorchTitan.
  - Torch Issue: https://github.com/pytorch/pytorch/issues/162638
  - TorchTitan implementation: https://github.com/pytorch/torchtitan/blob/b291ad662493b63d25b038a30a915082d3617baf/torchtitan/distributed/utils.py#L298
- Issue: Instead of calling forward() and backward() on the model and executing the loss_fct(), PP requires to call step() on the pipeline schedule.
  - Solution: Integrated scheduled pipeline into Trainer and Evaluator
- Note: Weight tying is currently not supported
- Question: Do we need to consider something regarding model init so that all corresponding model copies are initialized the same?
  - I.e.: How can we assure that parallel stages are initialized the same?
  - Other way around: Will stages containing similar model parts be initialized the same?
  - Answer: Since we first build that PP stages and then the FSDP2 parallelization and then init, this is fine.
  - Follow up question: What happens, if we use full replica data parallelization?
- Issue: Want to compare non-PP forward pass with PP forward pass in unit test
  - Goal: Compare losses
  - Solution: Added a non-PP config used to compute a comparison loss
  - Issue: Need non-PP forward pass loss on all ranks that contain a last stage of the PP forward pass.
    - Solution: Run a normal FSDP2 forward pass on all ranks and ignore the output on non-last stage ranks.
	- Remark: Running the FSDP2 pass only on the last stage ranks led to hanging. This might be a known issue with FSDP2.
  - Issue: Need the model to effectively compute the same forward pass.
    - Solution:
	  - In the test config, run model model initialization before sharding/staging (don't do this in a real training!).
	  - Fix torch seed before each initialization.
	  - Fix torch seed before generating each input sequence. Note that we thus have identical data parallel batches.
  - TODO: Numerical instability can still be observed and needs to be investigated further (in FSDP only runs as well).
- TODO: TP + PP
  - Issue: What is the correct sharding/staging order?
    - Solution: 1. PP 2. TP 3. FSDP
  - Issue: TP initialization modifies the layers of the model and runs into problems if those have been deleted for PP.
    - Solution: Adapted TP code to be able to handle missing/None layers.
- TODO?: MFU
- TODO: Flash attention not installable with PyTorch 2.9 nightly?
- TODO: Write more tests.
  - Backwards test using gradient norm to compare to normal FSDP2 run
- TODO: CLean up, in particular type hints and doc strings.

### DIFF
--- a/config_files/training/config_lorem_ipsum_long_fsdp2_pp.yaml
+++ b/config_files/training/config_lorem_ipsum_long_fsdp2_pp.yaml
@@ -24,7 +24,7 @@ settings:
     enforce_last_step_checkpointed: false
   step_profile:
     gradient_accumulation_steps: 1
-    local_train_micro_batch_size: 2
+    local_train_micro_batch_size: 4
     sequence_length: 256
   training_target:
     num_target_tokens:
@@ -222,7 +222,7 @@ scheduled_pipeline:
       pass_type: BY_REFERENCE
     pp_schedule_name: gpipe
     batch_size: ${settings.step_profile.local_train_micro_batch_size}
-    microbatch_size: 1
+    microbatch_size: 2
     pp_degree: ${device_mesh.config.pipeline_parallel_degree}
      # maybe better to use the fsdp model and the schedule here
      # instead of passing in the staged pipeline?
--- a/src/modalities/evaluator.py
+++ b/src/modalities/evaluator.py
@@ -36,6 +36,7 @@ class Evaluator:
         batch: DatasetBatch,
         model: nn.Module,
         loss_fun: Callable[[InferenceResultBatch], torch.Tensor],
+        scheduled_pipeline=None,  # TODO set type
     ) -> torch.Tensor:
         """Evaluate a single batch by forwarding it through the model and calculating the loss.
 
@@ -48,8 +49,26 @@ class Evaluator:
             torch.Tensor: The loss of the batch
         """
         with torch.no_grad():
-            result_batch = model_predict_batch(model=model, batch=batch)
-        loss = loss_fun(result_batch)
+            if scheduled_pipeline is not None:
+                pp_schedule = scheduled_pipeline.pp_schedule
+                targets, losses = (
+                    (batch.targets[loss_fun.target_key].contiguous(), [])
+                    if scheduled_pipeline.is_last_pp_stage
+                    else (None, None)
+                )
+
+                if scheduled_pipeline.is_first_pp_stage:
+                    pp_schedule.eval(batch.samples[model.sample_key].contiguous(), target=targets, losses=losses)
+                else:
+                    pp_schedule.eval(target=targets, losses=losses)
+                loss = (
+                    torch.mean(torch.stack(losses)).to(losses[0].device)
+                    if scheduled_pipeline.is_last_pp_stage
+                    else None
+                )
+            else:
+                result_batch = model_predict_batch(model=model, batch=batch)
+                loss = loss_fun(result_batch)
         return loss
 
     def evaluate(
@@ -58,6 +77,7 @@ class Evaluator:
         data_loaders: list[LLMDataLoader],
         loss_fun: Callable[[InferenceResultBatch], torch.Tensor],
         num_train_steps_done: int,
+        scheduled_pipeline=None,  # TODO set type
     ) -> dict[str, EvaluationResultBatch]:
         """Evaluate the model on a set of datasets.
 
@@ -90,10 +110,12 @@ class Evaluator:
                         batch=batch,
                         model=model,
                         loss_fun=loss_fun,
+                        scheduled_pipeline=scheduled_pipeline,
                     )
 
-                    cumulated_loss[0] += batch_loss.item()  # sum up batch loss
-                    cumulated_loss[1] += 1
+                    if batch_loss is not None:
+                        cumulated_loss[0] += batch_loss.item()  # sum up batch loss
+                        cumulated_loss[1] += 1
                     batch_length_tensor = torch.tensor(len(batch)).to(device)
                     thoughput_aggregator.add_value(key=ThroughputAggregationKeys.NUM_SAMPLES, value=batch_length_tensor)
 
--- a/src/modalities/gym.py
+++ b/src/modalities/gym.py
@@ -106,7 +106,7 @@ class Gym:
         evaluation_interval_in_steps: int,
         scheduled_pipeline=None,  # TODO set type
     ):
-        if num_train_steps_done % evaluation_interval_in_steps == 0 and num_train_steps_done > 10:
+        if num_train_steps_done % evaluation_interval_in_steps == 0:
             self.evaluator.evaluate(
                 model=model,
                 data_loaders=evaluation_data_loaders,
--- a/src/modalities/gym.py
+++ b/src/modalities/gym.py
@@ -106,7 +106,7 @@ class Gym:
         evaluation_interval_in_steps: int,
         scheduled_pipeline=None,  # TODO set type
     ):
-        if num_train_steps_done % evaluation_interval_in_steps == 0:
+        if num_train_steps_done > 0 and num_train_steps_done % evaluation_interval_in_steps == 0:
             self.evaluator.evaluate(
                 model=model,
                 data_loaders=evaluation_data_loaders,
--- a/src/modalities/gym.py
+++ b/src/modalities/gym.py
@@ -40,6 +40,7 @@ class Gym:
         train_data_loader: LLMDataLoader,
         evaluation_data_loaders: list[LLMDataLoader],
         checkpoint_saving: CheckpointSaving,
+        scheduled_pipeline=None,  # TODO set type
     ):
         """Runs the model training, including evaluation and checkpointing.
 
@@ -57,6 +58,7 @@ class Gym:
             model=app_state.model,
             evaluation_data_loaders=evaluation_data_loaders,
             evaluation_interval_in_steps=evaluation_interval_in_steps,
+            scheduled_pipeline=scheduled_pipeline,
         )
 
         checkpointing_callback: Callable[[TrainingProgress], None] = partial(
@@ -74,6 +76,7 @@ class Gym:
             evaluation_callback=evaluation_callback,
             checkpointing_callback=checkpointing_callback,
             training_log_interval_in_steps=training_log_interval_in_steps,
+            scheduled_pipeline=scheduled_pipeline,
         )
         print_rank_0(f"Training done at {datetime.now()}.")
 
@@ -101,11 +104,13 @@ class Gym:
         num_train_steps_done: int,
         evaluation_data_loaders: list[LLMDataLoader],
         evaluation_interval_in_steps: int,
+        scheduled_pipeline=None,  # TODO set type
     ):
-        if num_train_steps_done % evaluation_interval_in_steps == 0:
+        if num_train_steps_done % evaluation_interval_in_steps == 0 and num_train_steps_done > 10:
             self.evaluator.evaluate(
                 model=model,
                 data_loaders=evaluation_data_loaders,
                 loss_fun=self.loss_fun,
                 num_train_steps_done=num_train_steps_done,
+                scheduled_pipeline=scheduled_pipeline,
             )
--- a/src/modalities/loss_functions.py
+++ b/src/modalities/loss_functions.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from typing import overload
 
 import torch
 from torch.nn import CrossEntropyLoss
@@ -31,9 +32,16 @@ class CLMCrossEntropyLoss(Loss):
         # Mean over the tokens in the local-batch (batch per rank)
         self.loss_fun = CrossEntropyLoss(reduction="mean")
 
+    @overload
     def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
-        labels = forward_batch.get_targets(self.target_key)
-        lm_logits = forward_batch.get_predictions(self.prediction_key)
+        ...
+
+    @overload
+    def __call__(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        ...
+
+    def __call__(self, *args, **kwargs) -> torch.Tensor:
+        labels, lm_logits = self._parse_arguments(args, kwargs)
 
         # move labels to correct device to enable model parallelism
         labels = labels.to(lm_logits.device)
@@ -43,13 +51,40 @@ class CLMCrossEntropyLoss(Loss):
         loss = self.loss_fun(shift_logits.view(-1, shift_logits.size(-1)), shift_labels.view(-1))
         return loss
 
-
-class CLMCrossEntropyLossPP(CLMCrossEntropyLoss):
-    def __call__(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
-        forward_batch = InferenceResultBatch(
-            predictions={self.prediction_key: outputs}, targets={self.target_key: targets}
-        )
-        return super().__call__(forward_batch)
+    def _parse_arguments(
+        self,
+        args: list[torch.Tensor] | list[InferenceResultBatch],
+        kwargs: dict[str, torch.Tensor] | dict[str, InferenceResultBatch],
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        if len(args) == 1 and isinstance(args[0], InferenceResultBatch):
+            forward_batch = args[0]
+            labels = forward_batch.get_targets(self.target_key)
+            lm_logits = forward_batch.get_predictions(self.prediction_key)
+        elif "forward_batch" in kwargs and isinstance(kwargs["forward_batch"], InferenceResultBatch):
+            forward_batch = kwargs["forward_batch"]
+            labels = forward_batch.get_targets(self.target_key)
+            lm_logits = forward_batch.get_predictions(self.prediction_key)
+        elif len(args) == 2 and all(isinstance(arg, torch.Tensor) for arg in args):
+            lm_logits, labels = args
+        elif (
+            "outputs" in kwargs
+            and "targets" in kwargs
+            and isinstance(kwargs["outputs"], torch.Tensor)
+            and isinstance(kwargs["targets"], torch.Tensor)
+        ):
+            lm_logits = kwargs["outputs"]
+            labels = kwargs["targets"]
+        elif (
+            len(args) == 1
+            and "targets" in kwargs
+            and isinstance(args[0], torch.Tensor)
+            and isinstance(kwargs["targets"], torch.Tensor)
+        ):
+            lm_logits = args[0]
+            labels = kwargs["targets"]
+        else:
+            raise TypeError("Invalid arguments for CLMCrossEntropyLoss.__call__")
+        return labels, lm_logits
 
 
 def nce_loss(
--- a/src/modalities/loss_functions.py
+++ b/src/modalities/loss_functions.py
@@ -44,6 +44,14 @@ class CLMCrossEntropyLoss(Loss):
         return loss
 
 
+class CLMCrossEntropyLossPP(CLMCrossEntropyLoss):
+    def __call__(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        forward_batch = InferenceResultBatch(
+            predictions={self.prediction_key: outputs}, targets={self.target_key: targets}
+        )
+        return super().__call__(forward_batch)
+
+
 def nce_loss(
     embedding1: torch.Tensor, embedding2: torch.Tensor, device: torch.device, is_asymmetric: bool, temperature: float
 ) -> torch.Tensor:
--- a/src/modalities/models/gpt2/gpt2_model.py
+++ b/src/modalities/models/gpt2/gpt2_model.py
@@ -319,7 +319,7 @@ class GPT2LLMConfig(BaseModel):
         ffn_norm_config (LayerNormWrapperConfig): Config for normalization of the feed-forward network.
         lm_head_norm_config (LayerNormWrapperConfig): Config for normalization of the language model head.
         use_weight_tying (bool): Whether to use weight tying.
-
+        seed (int, optional): The seed for random number generation. Defaults to None.
     """
 
     sample_key: str
@@ -344,6 +344,7 @@ class GPT2LLMConfig(BaseModel):
     ffn_norm_config: LayerNormWrapperConfig
     lm_head_norm_config: LayerNormWrapperConfig
     use_weight_tying: bool
+    seed: Optional[int] = None
 
     @model_validator(mode="after")
     def check_divisibility(self) -> "GPT2LLMConfig":
--- a/src/modalities/models/gpt2/gpt2_model.py
+++ b/src/modalities/models/gpt2/gpt2_model.py
@@ -344,6 +344,7 @@ class GPT2LLMConfig(BaseModel):
     ffn_norm_config: LayerNormWrapperConfig
     lm_head_norm_config: LayerNormWrapperConfig
     use_weight_tying: bool
+    use_pp: Optional[bool] = False
 
     @model_validator(mode="after")
     def check_divisibility(self) -> "GPT2LLMConfig":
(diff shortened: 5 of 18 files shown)


src/modalities/gym.py
```diff
@@ -101,11 +104,13 @@ def _run_evaluation(
         num_train_steps_done: int,
         evaluation_data_loaders: list[LLMDataLoader],
         evaluation_interval_in_steps: int,
+        scheduled_pipeline=None,  # TODO set type
     ):
-        if num_train_steps_done % evaluation_interval_in_steps == 0:
+        if num_train_steps_done > 0 and num_train_steps_done % evaluation_interval_in_steps == 0:
```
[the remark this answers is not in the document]
rrutmann: Beause of a bug in pytorch, a model using pipeline parallelism first has to be used in training before evaluation

src/modalities/running_env/fsdp/device_mesh.py
```diff
@@ -127,3 +128,26 @@ def get_device_mesh(
...
+        return 1
+    else:
+        return device_mesh.size(device_mesh.mesh_dim_names.index(parallelism_method.value))
+
+
+def get_mesh_for_parallelism_method(device_mesh: DeviceMesh | None, parallelism_method: ParallelismDegrees):
+    if device_mesh is not None and parallelism_method.value in device_mesh.mesh_dim_names:
+        return device_mesh[parallelism_method.value]
+    else:
+        return None
```
le1nux: should we raise an exception instead?

src/modalities/trainer.py
```diff
@@ -106,19 +110,39 @@ def _train_batch(
...
+                else (None, None)
+            )
+
+            if scheduled_pipeline.is_first_pp_stage:
+                pp_schedule.step(batch.samples[model.sample_key].contiguous(), target=targets, losses=losses)
+            else:
+                pp_schedule.step(target=targets, losses=losses)
+            loss = torch.mean(torch.stack(losses)).to(losses[0].device) if scheduled_pipeline.is_last_pp_stage else None
+        else:
+            # else continue with loss calculation
```
le1nux: This code is very specific to pipeline parallelism. I would prefer if that would be abstracted away. However I also have no clear picture yet, how to do that. We should discuss this.

src/modalities/models/model_factory.py
```diff
@@ -631,7 +640,7 @@ def get_gpt2_tensor_parallelized_model(model: GPT2LLM, device_mesh: DeviceMesh)
             ),
         }
 
-        if isinstance(model.transformer.wpe, nn.Embedding):
+        if hasattr(model.transformer, "wpe") and isinstance(model.transformer.wpe, nn.Embedding):
```
le1nux: I checked the gpt2_model implementation. `model.transformer.wpe` should be always present either as an embedding or identity layer. Do we really need this check?

pyproject.toml
```diff
@@ -1,12 +1,10 @@
 [project]
 name = "modalities"
 version = "0.3.2"
```
le1nux: Why did we remove this? Our testing is always against 3.10 and 3.11. Do we need a more recent python version?

src/modalities/loss_functions.py
```diff
@@ -31,9 +32,16 @@ def __init__(self, target_key: str, prediction_key: str, tag: str = "CLMCrossEnt
...
-        labels = forward_batch.get_targets(self.target_key)
-        lm_logits = forward_batch.get_predictions(self.prediction_key)
+        ...
+
+    @overload
+    def __call__(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        ...
+
+    def __call__(self, *args, **kwargs) -> torch.Tensor:
+        labels, lm_logits = self._parse_arguments(args, kwargs)
```
le1nux: Could be improved from a software engineering point of view

src/modalities/loss_functions.py
```diff
@@ -43,6 +51,41 @@ def __call__(self, forward_batch: InferenceResultBatch) -> torch.Tensor:
...
+            and "targets" in kwargs
+            and isinstance(args[0], torch.Tensor)
+            and isinstance(kwargs["targets"], torch.Tensor)
+        ):
+            lm_logits = args[0]
+            labels = kwargs["targets"]
+        else:
+            raise TypeError("Invalid arguments for CLMCrossEntropyLoss.__call__")
+        return labels, lm_logits
+
```
le1nux: Idea: What about defining a new component "pp-loss", which takes a normal loss function and handles the PP-specific part?

src/modalities/evaluator.py
```diff
@@ -36,20 +37,42 @@ def evaluate_batch(
...
 
         Returns:
-            torch.Tensor: The loss of the batch
+            torch.Tensor | None: The loss of the batch
+                None, if a non-last stage was processed in pipeline parallelism
         """
         with torch.no_grad():
-            result_batch = model_predict_batch(model=model, batch=batch)
-        loss = loss_fun(result_batch)
+            if scheduled_pipeline is not None:
```
le1nux: basically code duplication from the trainer. 
Also not a big fan of passing the scheduled_pipeline in here.

le1nux (commented):
Awesome work with the PP integration! Functionality-wise everything looked correct (did not check the tests yet). 

Regarding the integration from a architectural perspetive, I left a couple of comments. I think we should do some refactorings here.

src/modalities/util.py
```diff
@@ -186,9 +186,11 @@ def get_total_number_of_trainable_parameters(model: FSDPX) -> Number:
         # However, users can also provide their own sharding process groups (currently not supported in Modalities)
         # which would require to adapt the code.
         if model.sharding_strategy.name == "NO_SHARD":
-            sharding_factor = dist.get_world_size()
+            sharding_factor = dist.get_world_size()  # TODO Check if we should use number of data parallel ranks instead
         if model.sharding_strategy.name == "HYBRID_SHARD":
-            sharding_factor = dist.get_world_size() // torch.cuda.device_count()
+            sharding_factor = (
+                dist.get_world_size() // torch.cuda.device_count()
+            )  # TODO Check if we should use number of data parallel ranks instead
```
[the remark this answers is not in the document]
rrutmann: Why should that be the case? Depending on the tp and pp degree, dp_degree could be different, right?

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -172,31 +183,43 @@ def clip_grad_norm_(
...
+        pp_mesh = get_mesh_for_parallelism_method(device_mesh=device_mesh, parallelism_method=ParallelismDegrees.PP)
+        if pp_mesh is not None:
+            if math.isinf(norm_type):
+                dist.all_reduce(total_norm, op=dist.ReduceOp.MAX, group=pp_mesh.get_group())
+            else:
+                total_norm **= norm_type
+                dist.all_reduce(total_norm, op=dist.ReduceOp.SUM, group=pp_mesh.get_group())
+                total_norm **= 1.0 / norm_type
+
+        torch.nn.utils.clip_grads_with_norm_(parameters, max_norm, total_norm, foreach)
```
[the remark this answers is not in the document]
BlueCrescent: Added.

config_files/training/config_lorem_ipsum_long_fsdp2_pp.yaml
```diff
@@ -222,7 +222,7 @@ scheduled_pipeline:
       pass_type: BY_REFERENCE
     pp_schedule_name: gpipe
     batch_size: ${settings.step_profile.local_train_micro_batch_size}
-    microbatch_size: 1
+    microbatch_size: 2
```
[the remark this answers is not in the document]
le1nux: Ok, for clarification:

we have the `local_train_microbatch_size` which is the batch size for a single rank for a single forward pass. The global batch size **per forward pass** (not per setp) would be `local_train_microbatch_size * dp_degree`. 
If we stage our model across `n` ranks, we run `n` microbatches in a single forward pass. Shouldn't `micro_batchsize = local_train_micro_batch_size // n` in this case?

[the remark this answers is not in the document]
rrutmann: The batch at each dp rank (with size local_train_microbatch_size) is split into microbatches of size micro_batchsize, which then pass through all pipeline stages one after the other. This size is initially independent of the number of pipeline stages.

config_files/training/config_lorem_ipsum_long_fsdp2_pp.yaml
```diff
@@ -0,0 +1,432 @@
...
+gradient_clipper:
+  component_key: gradient_clipper
+  variant_key: fsdp2
+  config:
+    wrapped_model:
+      instance_key: initialized_model
+      pass_type: BY_REFERENCE
+    norm_type: P2_NORM
+    max_norm: 1.0
+
```
le1nux: Don't we need a device mesh here?

src/modalities/models/gpt2/gpt2_model.py
```diff
@@ -819,7 +819,6 @@ def __init__(
             attention_norm_config (LayerNormWrapperConfig): Config for the attention normalization module.
             ffn_norm_config (LayerNormWrapperConfig): Config for the feed-forward network normalization module.
             lm_head_norm_config (LayerNormWrapperConfig): Config for the language model head normalization module.
-            seed (int, optional): The random seed. Defaults to None.
             use_weight_tying (bool): Whether to use weight tying.
```
le1nux: seed docstring missing

src/modalities/util.py
```diff
@@ -214,8 +217,13 @@ def get_total_number_of_trainable_parameters(model: FSDPX) -> Number:
         # >>> parameter_tensor.shape[0] * parameter_tensor.shape[1]
         # 6438912
 
-        total_num_params = sum(p.numel() for p in model.parameters() if p.requires_grad)
-        return total_num_params
+        num_params_tensor = sum(p.numel() for p in model.parameters() if p.requires_grad)
+        if device_mesh is not None and ParallelismDegrees.PP.value in device_mesh.mesh_dim_names:
```
le1nux: Do we have a test for this?

src/modalities/running_env/fsdp/device_mesh.py
```diff
@@ -127,3 +128,26 @@ def get_device_mesh(
...
+    Returns:
+        int: The number of parallel ranks for the specified parallelism method.
+    """
+    if parallelism_method.value not in device_mesh.mesh_dim_names:
+        return 1
+    else:
+        return device_mesh.size(device_mesh.mesh_dim_names.index(parallelism_method.value))
+
+
+def get_mesh_for_parallelism_method(device_mesh: DeviceMesh | None, parallelism_method: ParallelismDegrees):
```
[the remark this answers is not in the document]
le1nux: returning None in this case is also a bit unexpected.

src/modalities/trainer.py
```diff
@@ -250,14 +286,17 @@ def train(
...
+
+                cumulated_losses[1] = batch_loss.item() if batch_loss is not None else 0.0
 
                 reduced_losses = Reducer.reduce(
                     tensor=cumulated_losses,
                     operation=dist.ReduceOp.SUM,
                     # 1.) summed batch loss / (num batches * world size)
-                    # 2.) last batch loss / world size
-                    post_processing_fun=lambda t: torch.stack([t[0] / t[-1], t[1] / dist.get_world_size()]),
+                    # 2.) last batch loss / (world size / pp_degree)
```
le1nux: shouldn't it be
```
# 1.) summed batch loss / (num batches * dp_degree)
# 2.) last batch loss / dp_degree
```
?

I also don't really understand why we differentiate the two cases.

src/modalities/training/gradient_clipping/fsdp_gradient_clipper_config.py
```diff
@@ -26,7 +26,27 @@ class FSDPGradientClipperConfig(BaseModel):
     wrapped_model: PydanticPytorchModuleType
 
 
-class FSDPDummyGradientClipperConfig(BaseModel):
+class FSDP2GradientClipperConfig(FSDP1GradientClipperConfig):
```
le1nux: This is an "artificial" dependency in my opinion. I would just add
```
    max_norm: Annotated[float, Field(strict=True, gt=0)]
    norm_type: GradientClippingMode
    wrapped_model: PydanticPytorchModuleType
```
and remove the dependency.

src/modalities/training/gradient_clipping/fsdp_gradient_clipper_config.py
```diff
@@ -43,17 +63,21 @@ class FSDPDummyGradientClipperConfig(BaseModel):
     norm_type: GradientClippingMode
 
 
-class DummyGradientClipperConfig(BaseModel):
+class FSDP2DummyGradientClipperConfig(FSDP1DummyGradientClipperConfig):
```
le1nux: also here, I would resolve the dependency.

src/modalities/training/gradient_clipping/fsdp_gradient_clipper.py
```diff
@@ -88,65 +92,25 @@ def clip_gradients(self) -> torch.Tensor:
...
-        parameters: torch.Tensor | Iterable[torch.Tensor],
-        max_norm: float,
-        norm_type: float = 2.0,
+    def __init__(
+        self,
+        wrapped_model: FSDP2,
+        norm_type: GradientClippingMode,
+        device_mesh: Optional[DeviceMesh] = None,
         error_if_nonfinite: bool = False,
         foreach: Optional[bool] = None,
```
le1nux: are these to part of the config definition?

src/modalities/optimizers/optimizer_factory.py
```diff
@@ -156,14 +156,45 @@ def _create_optimizer_groups(
...
-            "weight_decay": weight_decay if group not in weight_decay_groups_excluded else 0.0,
+            "exclude": group not in weight_decay_groups_excluded,
         }
         for group in weight_decay_groups.keys()
     ]
-    return optimizer_groups, weight_decay_groups.keys()
+
+    optimizer_groups: OptimizerGroups = [
+        {
+            "params": sum((p["params"] for p in params_per_weight_decay_groups if not p["exclude"]), []),
```
le1nux: maybe add a commment that `sum` combines the parameters lists here. Was not aware that you could overload sum like this.

tests/fsdp2_parallelization/pipeline_parallelism/test_pp_fwd_bwd_pass.py
```diff
@@ -0,0 +1,166 @@
...
+    ) -> tuple[bool, torch.Tensor]:
+        try:
+            components = self._get_components(pp_model_config_path, use_pp=True)
+            scheduled_pipeline = components.scheduled_pipeline
+            loss_pp = self._forward_step(scheduled_pipeline, inputs, targets)
+        except Exception as e:
+            import traceback
+
+            print(f"Exception in _forward_step_with_pp: {e}")
+            traceback.print_exc()  # <-- Add this line to print the full stack trace
```
le1nux: *cough, cough ...* ChatGPT *cough, cough ...*

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
```diff
@@ -0,0 +1,276 @@
...
+        # saves multiple intermediary checkpoints.
+        # The second config starts from the 4th step and trains the model for 4 more steps.
+        # We compare the loss values of the two models after 4 steps and expect them to be the same.
+
+        try:
+            if tmp_folder.exists():
+                shutil.rmtree(tmp_folder)
+            tmp_folder.mkdir(parents=False, exist_ok=False)
+            # config for two steps model
+            gpt2_8_steps_config_file_path = working_dir / "gpt2_train_num_steps_7_pp_tp.yaml"
```
le1nux: 8 or 7 steps?

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
```diff
@@ -0,0 +1,276 @@
...
+
+            # adopt the checkpoint path
+            checkpoint_path = str(tmp_folder)
+            gpt2_8_steps_config_dict["checkpoint_saving"]["config"]["checkpoint_saving_execution"]["config"][
+                "checkpoint_path"
+            ] = checkpoint_path
+            gpt2_8_steps_config_dict["settings"]["paths"]["checkpoint_saving_path"] = checkpoint_path
+            loss_values_experiment_0_path = checkpoint_path + "/experiment_0_loss_scores.txt"
+
+            # config for one step model
```
le1nux: one or four step model?

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
```diff
@@ -0,0 +1,276 @@
...
+                        component_key="results_subscriber",
+                        variant_key="save_all",
+                        custom_component=SaveAllResultSubscriber,
+                        custom_config=SaveAllResultSubscriberConfig,
+                    )
+                    print(
+                        main_obj_0.config_dict["settings"]["training_target"]["num_target_tokens"]["config"][
+                            "dataset_path"
+                        ]
+                    )
```
le1nux: could be removed?

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
```diff
@@ -0,0 +1,276 @@
...
+            if int(os.environ.get("RANK", -1)) == 0:
+                try:
+                    if tmp_folder.exists():
+                        shutil.rmtree(tmp_folder)
+                except Exception as e:
+                    logging.warning(f"Rank {os.environ.get('RANK', -1)}: failed to remove tmp folder {tmp_folder}: {e}")
+
+    def test_warmstart_dataloader(self):
+        # non-skipped config
+        gpt2_two_steps_config_file_path = working_dir / "gpt2_train_num_steps_7_pp_tp.yaml"
```
le1nux: two or 7 steps?

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
```diff
@@ -0,0 +1,276 @@
...
+        finally:
+            logging.info(f"Rank {os.environ.get('RANK', -1)} cleaning up.")
+            if int(os.environ.get("RANK", -1)) == 0:
+                try:
+                    if tmp_folder.exists():
+                        shutil.rmtree(tmp_folder)
+                except Exception as e:
+                    logging.warning(f"Rank {os.environ.get('RANK', -1)}: failed to remove tmp folder {tmp_folder}: {e}")
+
+    def test_warmstart_dataloader(self):
```
le1nux: Do we have a test that checks if each rank sees the correct data in case of mutliple parallelization strategies?

le1nux (changes requested):
Amazing work! 🚀   I finally went through all of it and left a few comments (nothing major). 
It's already very solid. 

What do you think about one more end-to-end test that saves a FSDP+TP and FSDP+TP+PP checkpoint and continues the training from those checkpoints with FSDP and a different dp_degree (but compensating for the reduced token number by gradient accumulation). 

This would make sure that once we start the training, we can also switch hardware and parallelization strategies. I haven't seen such a test yet.

config_files/training/config_lorem_ipsum_long_fsdp2_pp_tp.yaml
```diff
@@ -0,0 +1,436 @@
...
+  variant_key: wandb
+  config:
+    global_rank: ${settings.cuda_env.global_rank}
+    project: modalities_dcp_tests
+    mode: OFFLINE
+    experiment_id: ${settings.experiment_id}
+    directory: wandb_storage
+    config_file_path: ${settings.config_file_path}
+
+# mfu_calculator:
```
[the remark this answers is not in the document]
rrutmann: We copied this from another config, where it was also commented out. Commented it in now

src/modalities/models/gpt2/gpt2_model.py
rrutmann: added the docstring again

tests/checkpointing/test_fsdp2_dcp_checkpoint_loading_and_saving.py
```diff
@@ -139,13 +184,21 @@ def _test_save_checkpoint_after_backward_pass_impl(
...
-        loss_0 = CheckpointingTestUtils.forward_backward_pass(
-            prediction_key=prediction_key,
-            model=app_state1.model,
-            optimizer=app_state1.optimizer,
-            batch_input_ids_dict=batch_input_ids_dict,
-            batch_target_ids=batch_target_ids,
-        )
+        if hasattr(app_state1, "scheduled_pipeline"):
+            loss_0 = CheckpointingTestUtils.forward_backward_pp_pass(
+                scheduled_pipeline=app_state1.scheduled_pipeline,
```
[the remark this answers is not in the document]
rrutmann: We added scheduled_pipeline explicitly to app_state1 in this test to avoid passing around scheduled_pipeline to other functions and keep the test more readable. Since the state_dict of app_state1 does not know about scheduled_pipeline, it is not checkpointed

src/modalities/util.py
rrutmann: We added a test for pipeline and tensor parallelism

src/modalities/util.py
```diff
@@ -162,12 +164,13 @@ def get_local_number_of_trainable_parameters(model: nn.Module) -> int:
...
 
 
-def get_total_number_of_trainable_parameters(model: FSDPX) -> Number:
+def get_total_number_of_trainable_parameters(model: FSDPX, device_mesh: DeviceMesh | None) -> Number:
     """Returns the total number of trainable parameters across all ranks.
     The model must be sharded with FSDP1 or FSDP2.
 
     Args:
         model (FSDPX): The model for which to calculate the number of trainable parameters.
+        device_mesh (DeviceMesh | None): The device mesh used for distributed training.
```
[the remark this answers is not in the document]
BlueCrescent: Added a corresponding assert for the FSDP2 case.

tests/fsdp2_parallelization/pipeline_parallelism/test_pp_fwd_bwd_pass.py
```diff
@@ -0,0 +1,166 @@
...
+            process_group_backend=ProcessGroupBackendType.nccl,
+            global_rank=process_id,
+            local_rank=process_id,
+            world_size=world_size,
+            rdvz_port=22359,
+        ):
+            vocab_size = 50304
+            sequence_length = 256
+            batch_size = 4
+            torch.manual_seed(42)
```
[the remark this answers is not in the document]
le1nux: If we created the data after this after the components instantiation:
https://github.com/Modalities/modalities/pull/399/files#diff-c6ccfa011be7a42ac0c0e6349e8d99d69cb6ad70780295522104a84428ccadf4R85-R89
we could make sure that each dp rank sees different data, as we could access the device mesh, no?

tests/end2end_tests/test_fsdp2_warmstart_pp_tp.py
BlueCrescent: Three additional tests have been added (to the dataloader unit tests).

src/modalities/util.py
```diff
@@ -197,6 +197,7 @@ def get_total_number_of_trainable_parameters(model: FSDPX, device_mesh: DeviceMe
         total_num_params = total_num_params // sharding_factor
         return total_num_params
     elif isinstance(model, FSDP2):
+        assert device_mesh is not None, "Device should be available for FSDP2 models."
```
le1nux: try except maybe?

tests/end2end_tests/configs/gpt2_warm_start_from_step_4_grad_accu.yaml
```diff
@@ -0,0 +1,328 @@
...
+      instance_key: collate_fn
+      pass_type: BY_REFERENCE
+
+eval_dataloaders: []
+
+# checkpoint_loading:
+#   component_key: checkpoint_loading
+#   variant_key: dcp
+#   config:
+#     global_rank: ${settings.cuda_env.global_rank}
```
le1nux: can we remove this

tests/end2end_tests/configs/gpt2_train_num_steps_7_grad_accu.yaml
```diff
@@ -0,0 +1,375 @@
...
+#           component_key: pipeline
+#           variant_key: selector
+#           config:
+#             pipeline:
+#               instance_key: staged_pipeline
+#               pass_type: BY_REFERENCE
+#             selection_type: PP_STAGE
+#         model_part:
+#           instance_key: fsdp_model
+#           pass_type: BY_REFERENCE
```
le1nux: legacy

tests/dataloader/distributed/test_distributed_multidim_dataloader.py
```diff
@@ -35,44 +42,144 @@ def test_distributed_multidim_dataloader_produces_different_data_on_different_dp
...
-            num_workers=2,
-            pin_memory=False,
-        )
+        device_mesh_mock = _build_device_mesh_mock(world_size, dp_degree, dp_rank, other_rank)
+        train_dataloader = _build_dataload_for_mesh(dataset_len, device_mesh_mock)
         return [batch.tolist() for batch in train_dataloader]
 
 
-def _build_device_mesh_mock(world_size: int, dp_degree: int, dp_rank: int, other_rank: int):
+def _build_dataload_for_mesh(dataset_len: int, device_mesh: DeviceMesh) -> LLMDataLoader:
```
le1nux: `_build_dataloader_for_mesh`

le1nux (approved):
LGTM :) I think we can merge this! 🥳 
Amazing work and loved the in-depth testing. 🚀

PR #426, opened 2025-12-10
fix: Diverse model seeding across PP ranks

rrutmann:
# What does this PR do?

This PR gives a unique model seed for each pp rank, such that parameters are initialized differently across ranks.

## General Changes
* On each rank, add the pp rank to the model seed.

## Breaking Changes
* None


BlueCrescent (commented):
Overall LGTM.
Should we also explicitly allow seeding for the "model_initialized" component?
It will probably inherit the random state from the model_raw component but it seems a bit risky to me to assume that (also in the future) no other interaction with the random state happens between these two components (though, probably, only interactions that are asymmetrical between the ranks would be problematic). In particular, since we cannot guarantee the order in which the components are build, something like a dataloader component might even re-seed the random state.

le1nux (changes requested):
I checked the seeding (not the test) and from my understanding the changes do not provide the expected results (also what @BlueCrescent  was hinting towards). 

When we seed the raw model, the model weights are indeed deterministic at instantiation time. However, we also have model weight initialization which runs afterwards and would just override the weights / seeding. 

Additionally, passing `device_mesh` to the model is coupling two components that should normally not know anything about each other. 

I think we have to integrate the seeding to the weight initializer component and can also pass in the `device_mesh` there.

rrutmann:
> I checked the seeding (not the test) and from my understanding the changes do not provide the expected results (also what @BlueCrescent was hinting towards).
> 
> When we seed the raw model, the model weights are indeed deterministic at instantiation time. However, we also have model weight initialization which runs afterwards and would just override the weights / seeding.
> 
> Additionally, passing `device_mesh` to the model is coupling two components that should normally not know anything about each other.
> 
> I think we have to integrate the seeding to the weight initializer component and can also pass in the `device_mesh` there.

Yes that makes sense. I moved the seeding to the model initialization component

rrutmann:
> Overall LGTM. Should we also explicitly allow seeding for the "model_initialized" component? It will probably inherit the random state from the model_raw component but it seems a bit risky to me to assume that (also in the future) no other interaction with the random state happens between these two components (though, probably, only interactions that are asymmetrical between the ranks would be problematic). In particular, since we cannot guarantee the order in which the components are build, something like a dataloader component might even re-seed the random state.

See https://github.com/Modalities/modalities/pull/426#issuecomment-3675521456

src/modalities/models/model.py
```diff
@@ -26,7 +26,7 @@ class ActivationType(str, Enum):
 class NNModel(nn.Module):
     """NNModel class to define a base model."""
 
-    def __init__(self, seed: int = None, weight_decay_groups: Optional[WeightDecayGroups] = None):
+    def __init__(self, seed: Optional[int] = None, weight_decay_groups: Optional[WeightDecayGroups] = None):
```
le1nux:
```suggestion
    def __init__(self, seed: int | None = None, weight_decay_groups: Optional[WeightDecayGroups] = None):
```

le1nux: Do we even want to allow setting the seed here? 
Could `torch.manual_seed` below have side effects with the new weight init implementation?

src/modalities/nn/model_initialization/composed_initialization.py
```diff
@@ -30,6 +32,8 @@ class ComposedModelInitializationConfig(BaseModel):
     std: Annotated[float, Field(strict=True, ge=0.0)] | str  # can be float or "auto"
     hidden_dim: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
     num_layers: Optional[Annotated[int, Field(strict=True, gt=0)]] = None
+    seed: Optional[int] = None
```
le1nux:
```suggestion
    seed: int | None = None
```

le1nux: also other instances and docstring annotations should be adapted accordingly

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -109,20 +119,29 @@ def get_scaled_initialization(
...
         """
+        InitializationRoutines._set_seed(seed)
         std = math.sqrt(0.4)
         initialization = NamedParameterwiseNormalInitialization(
             mean=mean, std=std, parameter_name_regexes=parameter_name_regexes
         )
         return initialization
+
+    @staticmethod
+    def _set_seed(seed: Optional[int]):
```
le1nux: this sets the seed globally. I think an even more robust way would be to use a local torch rng object. 
Could this be integrated? 

Something like: 
```py
g = torch.Generator()
g.manual_seed(1234)
```

BlueCrescent (approved):
LGTM

src/modalities/nn/model_initialization/composed_initialization.py
```diff
@@ -114,16 +120,28 @@ def get_composed_model_initializer(
...
+            seed (Optional[int], optional): Seed for random initialization. Defaults to None.
 
         Returns:
             ModelInitializationIF: The Weight Initializer performing the initialization as specified.
         """
+        # Set different random seed for each PP rank to ensure diversity
+        if seed is not None and has_parallelism_method(
+            device_mesh=device_mesh, parallelism_method=ParallelismDegrees.PP
+        ):
+            seed += get_parallel_rank(device_mesh=device_mesh, parallelism_method=ParallelismDegrees.PP)
```
le1nux, line 169: This also means that depending on the parallelization method and also the number of parallelism degrees we get differently initialized layers even if the seed is the same.
Example: 

DP with seed = 1, will have a differently initilized model than DP+PP with seed = 1. 

One way to fix this is to always use the same seed but each PP stage has to skip the number of random values of the pervious stages. 
However, I think this would be overkill and I would just place a warning when initialising the weights and parallelization methods are other than FSDP.
What do you think?

le1nux: We could also give the hint that for full reproducibility a Distributed Checkpoint with FSDP directly after weight init. 
Maybe we could even have an entry point for that in main. 

something like:
`modalities create_init_cp model_config.yaml`

For some unit tests, this functionality would be nice to have anyways I think.

Any thoughts?

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -62,7 +63,11 @@ def initialize_in_place(self, model: nn.Module):
 class InitializationRoutines:
     @staticmethod
     def get_plain_initialization(
-        mean: float, std: float | str, parameter_name_regexes: list[str], hidden_dim: Optional[int] = None
+        mean: float,
+        std: float | str,
+        parameter_name_regexes: list[str],
+        hidden_dim: Optional[int] = None,
+        seed: Optional[int] = None,
```
le1nux:
```suggestion
        seed: int | None = None,
```

also various cases below

src/modalities/nn/model_initialization/initialization_routines.py
le1nux: Otherwise, we might get into side-effects later on

le1nux (changes requested):
Generally good state. 
Left a couple of comments. 
My main concern is the global setting of the seed. A generator object might be favorable.

src/modalities/models/model.py
rrutmann: Probably it could have side effects, e.g. default submodule initialization, random ops and the ambient global RNG state for unrelated code. Also it is mostly redundant since we now use a local generator for weight initialization. I would suggest to remove it.

src/modalities/nn/model_initialization/initialization_routines.py
rrutmann: Makes sense. I integrated this

src/modalities/nn/model_initialization/composed_initialization.py
rrutmann: Good catch. I added a warning for now, but the additional entry point would be nice to have as well. I created an [issue](https://github.com/Modalities/modalities/issues/443) for that

src/modalities/conversion/gpt2/modeling_gpt2.py
```diff
@@ -40,7 +40,14 @@
 from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
 from transformers.processing_utils import Unpack
 from transformers.utils import TransformersKwargs, auto_docstring, can_return_tuple, logging
-from transformers.utils.generic import check_model_inputs
+
+try:
+    from transformers.utils.generic import check_model_inputs
+except ImportError:
+
+    def check_model_inputs(func: Callable) -> Callable:
```
le1nux, line 48: was this removed in transormers? 
If it is part of a legacy API I think we should also remove this on our end. 
What do you think @BlueCrescent? I think you added it, right?

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -39,21 +40,32 @@ class ScaledEmbedInitializationConfig(BaseModel):
...
 class NamedParameterwiseNormalInitialization(ModelInitializationIF):
-    def __init__(self, mean: float, std: float, parameter_name_regexes: RegexFilter):
+    def __init__(self, mean: float, std: float, parameter_name_regexes: RegexFilter, seed: int | None = None):
         self.mean = mean
         self.std = std
         self.parameter_name_regexes = parameter_name_regexes
+        self.seed = torch.initial_seed() if seed is None else seed
+        self._generators: dict[str, torch.Generator] = {}
+
+    def _get_generator(self, parameter: torch.Tensor) -> torch.Generator:
```
le1nux, line 66: a few things are not clear to me. 
 1. Do we actually have the case, where in a single process tensors are sitting on different GPUs? 
 2. if 1. is the case, then we can end up with tensors that are initialized identically, since we create multiple generators from the same seed. 

I'm not sure what the best way to solve this ... also seems to me that the Pytorch API regarding Generators is kinda limited.

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -73,23 +89,26 @@ def get_plain_initialization(
             std (float): standard deviation of the normal distribution. If set to "auto", appropiate
                 value selected as per plain initialization described in https://arxiv.org/abs/2312.16903
             hidden_dim (Optional[int]): hidden dimension of the attention layer. Defaults to None.
+            parameter_name_regexes (list[str]): List of parameter name regexes to which the initialization
```
le1nux, line 118: I think the type annotation is wrong. should be `RegexFilter`

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -99,6 +118,7 @@ def get_scaled_initialization(
             num_layers (int): Number of layers in the model which we use to downscale std with
             parameter_name_regexes (list[str]): List of parameter name regexes to which the initialization
```
le1nux:
```suggestion
            parameter_name_regexes (RegexFilter): List of parameter name regexes to which the initialization
```

le1nux (changes requested):
Looks good to me! Nice work :) 
The comment regarding the per-device Generator we should discuss, what makes most sense here. 

I would add one last test, which checks that two models instantiated with the same config file (with a specified seed), should have 100% matching parameter weights. I'd keep that one simple (no advanced sharding like TP or PP. only FSDP).

src/modalities/conversion/gpt2/modeling_gpt2.py
rrutmann: The function was removed in transformers version 5.2. In our pyproject.yaml we specify the requirement  "transformers>=4.57.4,<5.0.0", so I used an unsupported transformers version here. Should we remove it just to be on the safe side?

src/modalities/nn/model_initialization/initialization_routines.py
rrutmann: Since we start a single process for each rank via torchrun, this shouldn't happen, right? Or do I miss something?

src/modalities/conversion/gpt2/modeling_gpt2.py
BlueCrescent: Yes, I think, we should tackle the transformers 5.0.0+ support soon anyways.

src/modalities/nn/model_initialization/initialization_routines.py
```diff
@@ -39,21 +41,49 @@ class ScaledEmbedInitializationConfig(BaseModel):
...
 
 class NamedParameterwiseNormalInitialization(ModelInitializationIF):
-    def __init__(self, mean: float, std: float, parameter_name_regexes: RegexFilter):
+    def __init__(
+        self,
+        mean: float,
+        std: float,
+        parameter_name_regexes: RegexFilter,
+        seed: int | None = None,
+        multi_device_generator_policy: Literal["ignore", "warn", "error"] = "warn",
```
le1nux: Since we instantiate this from outside, I would make this an enum

le1nux (approved):
Minor comment. Otherwise all looks good to me! :) 
Nice work! 👍

PR #455, opened 2026-06-19
Add automatic Downstream Evaluation to Modalities - Simplified Implementation

CYHSM:
# What does this PR do?

This PR is the bare version of https://github.com/Modalities/modalities/pull/450 which works with a converted HF checkpoint. For this to work we need to first merge the conversion code from https://github.com/Modalities/modalities/pull/424 and then merge this on top. 

In https://github.com/Modalities/modalities/pull/450 it integrates the conversion + the evaluation but @le1nux suggested to first merge @BlueCrescent PR. For more details see original PR. 

### DIFF
--- a/src/modalities/config/instantiation_models.py
+++ b/src/modalities/config/instantiation_models.py
@@ -22,6 +22,7 @@ from modalities.config.pydantic_if_types import (
     PydanticSteppableProfilerIFType,
     PydanticTextInferenceComponentType,
     PydanticTokenizerIFType,
+    PydanticDownstreamEvaluatorType,
 )
 from modalities.config.utils import parse_torch_device
 from modalities.dataloader.dataset import Dataset
@@ -192,6 +193,7 @@ class TrainingComponentsInstantiationModel(BaseModel):
     mfu_calculator: PydanticMFUCalculatorABCType | None = None
     scheduled_pipeline: PydanticPipelineType | None = None
     device_mesh: PydanticDeviceMeshIFType | None = None
+    downstream_evaluator: Optional[PydanticDownstreamEvaluatorType] = None
     model_raw: PydanticPytorchModuleType
 
     @model_validator(mode="after")
--- a/src/modalities/evaluator.py
+++ b/src/modalities/evaluator.py
@@ -1,9 +1,15 @@
 from typing import Callable
+import json
+import logging
+import subprocess
+from pathlib import Path
 
 import torch
 import torch.distributed as dist
 import torch.nn as nn
 from torch.distributed.device_mesh import DeviceMesh
+
+from modalities.tokenization.tokenizer_wrapper import TokenizerWrapper
 
 from modalities.batch import DatasetBatch, EvaluationResultBatch, InferenceResultBatch, ResultItem
 from modalities.dataloader.dataloader import LLMDataLoader
@@ -14,6 +20,8 @@ from modalities.models.parallelism.pipeline_parallelism import Pipeline
 from modalities.running_env.fsdp.device_mesh import ParallelismDegrees, get_parallel_degree
 from modalities.running_env.fsdp.reducer import Reducer
 from modalities.util import TimeRecorder
+
+logger = logging.getLogger(__name__)
 
 
 class Evaluator:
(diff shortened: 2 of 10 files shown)


src/modalities/config/instantiation_models.py
```diff
@@ -192,6 +193,7 @@ def _check_last_step_checkpointed(self) -> "TrainingComponentsInstantiationModel
     mfu_calculator: PydanticMFUCalculatorABCType | None = None
     scheduled_pipeline: PydanticPipelineType | None = None
     device_mesh: PydanticDeviceMeshIFType | None = None
+    downstream_evaluator: Optional[PydanticDownstreamEvaluatorType] = None
```
le1nux, line 196:
```suggestion
    downstream_evaluator: PydanticDownstreamEvaluatorType | None = None
```

src/modalities/evaluator.py
```diff
@@ -197,3 +205,135 @@ def _publish_evaluation_result(
         evaluation_result_publisher.publish_message(
             payload=evaluation_result, message_type=MessageTypes.EVALUATION_RESULT
         )
+
+
+class DownstreamEvaluator:
```
le1nux, line 210: maybe we should have this implement an EvaluationIF interface

src/modalities/evaluator.py
```diff
@@ -197,3 +205,135 @@ def _publish_evaluation_result(
...
+        if hf_model_dir is None:
+            logger.warning(
+                f"No hf_checkpoint found in {self.checkpoint_dir} at step {num_train_steps_done}, "
+                "skipping downstream evaluation."
+            )
+            return
+
+        tasks_str = " ".join(self.tasks)
+        cmd = self.olmes_command_template.format(
+            hf_model_dir=str(hf_model_dir),
```
le1nux, line 251: I think running eval on the same node(s) as the training is unrealistic. Typically especially for larger models GPU memory headroom for evaluation is rather limited.

src/modalities/evaluator.py
```diff
@@ -197,3 +205,135 @@ def _publish_evaluation_result(
...
+        logger.info(f"Running downstream evaluation: {cmd}")
+        try:
+            p = subprocess.Popen(cmd, shell=True)
+            self.active_processes.append((p, num_train_steps_done, hf_model_dir))
+            logger.info(f"Downstream evaluation launched for step {num_train_steps_done}.")
+        except Exception as e:
+            logger.error(f"Failed to launch downstream evaluation: {e}")
+
+    def wait_for_evaluations(self) -> None:
+        if not hasattr(self, "active_processes") or not self.active_processes:
```
le1nux, line 265: why so complicated? It's an instance variable. 

```suggestion
        if len(self.active_processes) == 0:
```

src/modalities/main.py
```diff
@@ -229,8 +233,16 @@ def run(self, components: TrainingComponentsInstantiationModel):
...
+            downstream_evaluation_callback=downstream_evaluation_callback,
         )
 
+        if getattr(components, "downstream_evaluator", None) is not None:
+            print_rank_0("\n" + "="*80)
+            print_rank_0("Training loop complete! Waiting for background evaluations to finish...")
+            print_rank_0("="*80 + "\n")
+            components.downstream_evaluator.wait_for_evaluations()
+            print_rank_0("All background evaluations completed successfully!")
+
```
le1nux, line 245: as written in the overall feedback, I think we should exclude evaluation from the eval loop.

le1nux (changes requested):
I think this is a good draft but we need to consider a few architectural / workflow adaptations for large-scale setups. 
When we train large scale models the required training GPUs exceed the required eval GPUs by far. Also the training typically utilized most of the GPU memory, leaving no memory for evals potentially. 

That's why I would not include the downstream eval into the training / eval loop of modalities and decouple this instead.

HF checkpoint conversion and downstream eval should run on dedicated, non-training nodes as new checkpoints are written out during the training. 

So a better way in my opinion would be to have a CP conversion and eval service running in the background, polling  e.g., every 60s for new checkpoints including explicit sentinel. 

CP conversion and downstream eval can be CMD entry points within modalities.

PR #459, opened 2026-07-28
feat(model): add Nemotron-3 Nano hybrid Mamba-Transformer MoE architecture

le1nux:
Adds support for hybrid Mamba-Transformer architectures with sparse mixture-of-experts feed-forward layers, targeting **Nemotron-3 Nano 30B-A3B** ([arXiv:2512.20848](https://arxiv.org/abs/2512.20848)). Reference behaviour was taken from Megatron-Bridge's `HybridModelProvider` recipe and Megatron-LM's `megatron/core/ssm` + `megatron/core/models/hybrid`.

**~2,900 lines of implementation, ~3,100 lines of tests, 258 tests.** The GPT-2 implementation is untouched.

## What the architecture is, and why it needs new components

I confirmed the spec from the model report (Table 1 / Figure 2) and cross-checked it against the reference recipe; they agree. Three properties drove nearly every design decision:

**1. Layers are single-operator, not blocks.** A classical transformer block bundles attention and an MLP. Here every layer wraps exactly *one* operator in a pre-norm residual, and the stack is described by a pattern string:

```
MEMEM*EMEMEM*EMEMEM*EMEMEM*EMEMEM*EMEMEMEM*EMEMEMEME   # 52 layers
```

23 Mamba-2 (`M`), 23 MoE (`E`), and only **6** attention (`*`). Because the Mamba-2 layers carry positional information, the model uses **no positional embeddings at all**. Changing the Mamba/attention/MoE ratio is a one-line config edit.

**2. Attention decouples head dim from model dim.** 32 query heads × 128 head dim = 4096, on a model dimension of 2688. GPT-2's attention hard-derives `head_dim = n_embd // n_head_q`, so it genuinely cannot express this — a new module was required, not a preference.

**3. Experts are non-gated.** Squared ReLU, so two matrices per expert instead of three. My parameter accounting reproduces both published figures exactly (31.6B total, 3.2B active / 3.6B with embeddings), which is what confirmed the non-gated reading and that Megatron fuses the 2 shared experts into one MLP of hidden 3712 = 2×1856.

## The central design decision: layer *specs*, not layer instances

`ComponentFactory` memoises every config node. Injecting an instantiated `mamba_layer` component would have made **all 23 Mamba layers share one weight tensor** — a model that trains and silently underfits. This is exactly what Megatron-LM's `ModuleSpec` exists to avoid.

So each network component is registered as a **builder**:

```python
class NemotronLayerSpecIF(ABC):
    @property
    def symbol(self) -> LayerSymbol: ...   # M | * | E | -
    def build(self, layer_idx: int) -> nn.Module: ...
```

The model calls `build()` once per position. Every hyperparameter of every component stays reachable from YAML, while each layer gets independent parameters. `test_repeated_layer_types_have_independent_weights` pins this down through the real YAML path.

## Structural choice: mirror the GPT-2 module tree

```
transformer.{wte, h.<i>, lm_head_norm, lm_head}
```

`transformer.h` is an `nn.ModuleDict` because that is what the activation-checkpointing component requires. The payoff: `activation_checkpointed`, `fsdp2_wrapped`, pipeline splitting, and `ChunkedCLMCrossEntropyLoss` all work **unchanged** — no new wrapper components, and `layers_fqn: transformer.h` just works.

## Load balancing

Nemotron uses two mechanisms and both are implemented:

- **Aux-loss-free expert bias** (primary, rate 1e-3) as an **optimizer step pre-hook**, registered by a new `optimizer` decorator variant `moe_load_balanced`. Attaching it to the optimizer step rather than the forward pass is what makes it correct under gradient accumulation: counts accumulate over micro-batches and reduce across data-parallel ranks exactly once per step. Because the rule uses only `sign(mean − load)`, activation-checkpointing's double-counted forward is harmless.
- **Classic load-balancing loss** (secondary, 1e-4), computed per sequence, surfaced through the output dict and combined via a new `weighted_sum` loss. Logits are inserted first so `InferenceResultBatch.__len__` stays correct.

## Four real bugs found and fixed during the work

These are the parts worth reviewing most closely — each would have been silent in production:

1. **`GroupedExperts` weights were uninitialized memory.** `w1`/`w2` are raw `nn.Parameter(torch.empty(...))`, so nothing initialized them if a regex failed to match. Found via a NaN that only appeared under a specific test ordering. Fixed with a `reset_parameters` giving `nn.Linear`-equivalent init.

2. **`reset_parameters` wrote plain tensors into DTensors.** Under FSDP2, `dt_bias.copy_(plain_tensor)` raises. Fixed by writing into the local shard — valid because the values are i.i.d. per head. Found only by the distributed test.

3. **The pipeline stage packer silently dropped the `lm_head`.** The inherited greedy packer assumes uniform layer weights; with per-layer-type costs it exhausts its budget early and drops trailing modules. Verified directly:
   ```
   MISSING: {'transformer.lm_head_norm', 'transformer.lm_head'}
   ```
   Replaced with a binary-search-plus-split partitioner that provably covers every module. The 52-layer model now splits into 4 stages with weights `[31, 32, 31, 31]` (ideal 31.25).

4. **Weight init silently did nothing whenever activation checkpointing was enabled.** The shared initializer stripped only torch.compile's `_orig_mod.`, not `_checkpoint_wrapped_module.`, so **every per-layer regex failed to match** and models kept their default init. **This affected GPT-2 configs identically.** Fixed in `initialization_routines.py` by normalizing all wrapper prefixes. This is the one change made outside the stated scope — leaving a known silent-correctness bug in the path this feature's own config uses would have been wrong. `Llama3Initializer` has the same gap but lives in the GPT-2 package, so it was left alone and documented.

## Testing

| Suite | Result |
|---|---|
| Nemotron + components (CPU/GPU) | 253 passed, 3 skipped |
| Distributed FSDP2 (2 GPUs) | 2 passed |
| `tests/models`, `nn`, `config`, optimizer, loss, weight-tying | 362 passed, 4 skipped |
| Full collection | 814 tests collect cleanly |
| isort / black / ruff | clean |

The correctness anchor is `test_ssd.py`: the chunked SSD scan is validated against a step-by-step transcription of the Mamba-2 recurrence — not against itself — plus chunk-size invariance, causality, state-passing across a split sequence, and gradient flow. The fused Triton path is then checked against the native one, so it inherits that validation transitively.

Causality is asserted at every level (scan, conv, mixer, attention, assembled model), because a causality bug leaks the target and shows up as a suspiciously *good* loss rather than a crash.

**Not verified:** the `end2end_tests` and checkpointing suites need a multi-GPU launcher. Note also that `TestLlama3LikeInitialization` is **pre-existing flaky** — baselined with this PR's changes reverted, it still fails, with a different parametrization each run.

## Caveats

**The default Mamba backend is slow.** `ssd_backend: native` is dependency-free, CPU-runnable and `torch.compile`-friendly, but materially slower and more activation-memory hungry than the Triton kernels. It exists so the model is testable without `mamba-ssm` (not a Modalities dependency). For real 30B training you want `pip install -e '.[mamba]'` and `ssd_backend: fused` — the fused parity tests were skipped in this environment, so that path is written but **unverified here**.

**No tensor, expert, or context parallelism.** Mamba TP is genuinely hard: `in_proj` is a five-way unequal column split, and `conv1d`/`A_log`/`D`/`dt_bias` need per-head sharding with partition metadata. Expert parallelism has no dimension in the device mesh at all. The reference recipe trains 30B-A3B with EP=8; here you get FSDP2 + AC + PP, which is enough to train the model but not at the reference topology or throughput.

**MoE aux loss and pipeline parallelism don't compose.** Under PP, MoE layers on non-final stages can't reach the loss. The aux-loss-free bias balancing is layer-local and unaffected, which is why it is the primary mechanism — but if you enable PP, drop the `weighted_sum` loss term.

**Numerics are unvalidated against the reference.** Parameter shapes, initialization distributions and the routing formula were matched to Megatron-LM by reading the source, and the parameter counts reproduce the paper. No Nemotron checkpoint was loaded to compare logits. Until someone does, treat architectural equivalence as "carefully derived", not "measured".

**MFU is an estimate.** The formula charges the quadratic term only to the 6 attention layers and uses active parameters, but it ignores the SSM scan's own FLOPs, so it will read slightly optimistic.

## Next steps for production at scale

1. **Numerical equivalence against the reference.** Convert an HF `NemotronHForCausalLM` checkpoint and assert logit parity. The parameter mapping is fully specified in `nemotron_h_bridge.py:369-410`, and module names were chosen to keep that mapping direct. This is the single highest-value follow-up — everything else is performance.
2. **Fused backend validation and benchmarking.** Install the extra, run the parity tests on GPU, then measure tokens/s and peak memory against the native path to size the real gap.
3. **Tensor parallelism for Mamba.** Shard `in_proj` with explicit per-component partition sizes and `conv1d`/`A_log`/`D`/`dt_bias` per head. Attention and MoE TP are straightforward by comparison.
4. **Expert parallelism.** Requires an `ep` dimension in `DeviceMeshConfig` plus an all-to-all token dispatcher. This is the difference between training 30B-A3B on 8 GPUs and on 512. TorchTitan's `token_dispatcher.py` is a reasonable template.
5. **Selective-op activation checkpointing tuning.** The reference recomputes selectively over `["moe", "layernorm"]`; the `SELECTIVE_OP` variant can express something similar, but the op list needs tuning for the SSD scan.
6. **MTP and long-context.** Multi-token prediction (`mtp_hybrid_override_pattern="*E"`) and the 512k-sequence long-context phase are both absent; the latter needs context parallelism with Mamba state passing.
7. **Convergence smoke test.** Train the tiny variant to a known loss on lorem-ipsum, then a ~1B config on real data for a few thousand steps, watching expert-load entropy to confirm the bias balancing actually prevents collapse at scale. The unit tests prove the update rule is correct; only a real run proves it is *sufficient*.

## Docs

See [`docs/components/nemotron.md`](docs/components/nemotron.md) for the full guide, and [`config_files/training/config_nemotron3_nano_30b_a3b_fsdp2.yaml`](config_files/training/config_nemotron3_nano_30b_a3b_fsdp2.yaml) for a ready-to-run 30B-A3B pretraining config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -4,6 +4,7 @@
 |------------------|------------|---------------|------------------|------------------------------------------------------------------------------------------------|
 | [#141](#pr-141-towards-stable-modalities-version)  | Bug Fix    |  [#129](https://github.com/Modalities/modalities/issues/129)         | **Yes**              | Towards stable modalities version                                                               |
 | [#154](pr-154-manual-swiglu-implementation)  | Bug Fix    |  [#14](https://github.com/Modalities/modalities/issues/14)         | **Yes**              | Towards stable modalities version                                                               |
+| [#nemotron](#pr-nemotron-3-nano-hybrid-mamba-transformer-moe-architecture)  | Feature    |  --         | No               | Nemotron-3 Nano hybrid Mamba-Transformer MoE architecture                                      |
 |    |   |           |        |                                                                |
 
 
@@ -218,3 +219,53 @@ This PR improves training monitoring and logging across runs besides some other 
 
 **Breaking Changes**
 * experiments_root_path is now exposed on an API level
+
+## PR Nemotron 3 Nano hybrid Mamba-Transformer MoE architecture
+
+Adds support for hybrid Mamba-Transformer architectures with sparse mixture-of-experts feed-forward
+layers, targeting **Nemotron-3 Nano 30B-A3B** ([arXiv:2512.20848](https://arxiv.org/abs/2512.20848)).
+See [docs/components/nemotron.md](docs/components/nemotron.md) for the full guide.
+
+**General changes:**
+* New `model` variant `nemotron`: a decoder-only LM whose layer stack is described by a *layer
+  pattern* string (`M` = Mamba-2, `*` = attention, `E` = MoE, `-` = dense MLP), where each character
+  is one single-operator pre-norm residual layer. No positional embeddings; the Mamba-2 layers carry
+  the positional information.
+* New `nemotron_layer_spec` component type with the variants `mamba2`, `attention`, `moe` and `mlp`.
+  These are *builders*, not modules: the model calls `build()` once per layer position, so repeated
+  layer types get independent weights while every hyperparameter stays configurable from YAML.
+* New reusable components under `models/components/`: a Mamba-2 mixer with a dependency-free
+  pure-PyTorch state space scan (`ssd_backend: native`) plus an optional fused Triton backend
+  (`ssd_backend: fused`, new `mamba` extra), and a mixture-of-experts block with a sigmoid top-k
+  router, grouped-matmul experts and shared experts.
+* New `optimizer` variant `moe_load_balanced`: an optimizer decorator implementing auxiliary-loss-free
+  load balancing ([arXiv:2408.15664](https://arxiv.org/abs/2408.15664)) as a step pre-hook, which
+  makes the expert bias update correct under gradient accumulation.
+* New `loss` variants `moe_aux_loss` and `weighted_sum`, so the sequence-level MoE load-balancing
+  penalty can be added to the language modelling objective.
+* New `stages_generator` variant `nemotron_stages_generator`, which balances pipeline stages by
+  per-layer-type computational cost instead of layer count, and guarantees that every module is
+  assigned to exactly one stage.
+* New `mfu_calculator` variant `nemotron`, which uses *active* parameters and charges the quadratic
+  attention term only to the attention layers.
+* New `model_initialization` model type `nemotron`. The state space parameters (`A_log`, `D`,
+  `dt_bias`, `conv1d_*`) are excluded from the regex-driven initializer and initialized by
+  `Mamba2Mixer.reset_parameters` instead, matching the reference distributions.
+* Ready-to-run config: `config_files/training/config_nemotron3_nano_30b_a3b_fsdp2.yaml`.
+
+**Bug fixes (pre-existing, surfaced by the distributed tests for this feature):**
+* `NamedParameterwiseNormalInitialization` only stripped torch.compile's `_orig_mod.` prefix from
+  parameter names before matching the initialization filters. Activation checkpointing and FSDP1
+  insert their own segments (`_checkpoint_wrapped_module.`, `_fsdp_wrapped_module.`), so any config
+  that applied activation checkpointing before weight initialization silently matched *no* per-layer
+  regex and left the model with its default initialization. All wrapper prefixes are now normalized
+  away. This affected GPT2 configs as well. Note that `Llama3Initializer` (in the GPT2 package) still
+  has the same gap and was deliberately left untouched.
+
+**Breaking changes:**
+* None. The GPT2 implementation is untouched.
+
+**Known limitations:**
+* Tensor, expert and context parallelism are not supported for this architecture yet. FSDP2,
+  activation checkpointing, pipeline parallelism and `torch.compile` are.
+* Multi-token prediction (MTP) and HuggingFace checkpoint conversion are not included.
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -253,7 +253,15 @@ See [docs/components/nemotron.md](docs/components/nemotron.md) for the full guid
   `Mamba2Mixer.reset_parameters` instead, matching the reference distributions.
 * Ready-to-run config: `config_files/training/config_nemotron3_nano_30b_a3b_fsdp2.yaml`.
 
-**Bug fixes (pre-existing, surfaced by the distributed tests for this feature):**
+**Bug fixes (pre-existing, surfaced while building and running this feature):**
+* `DCPCheckpointSaving._delete_checkpoint` removed checkpoints with `Path.rmdir()`, which only works
+  on *empty* directories. A distributed checkpoint is a directory of per-rank shard files, so
+  deleting one always raised `OSError("Directory not empty")`. Every rotating checkpoint strategy
+  (`save_k_most_recent_checkpoints_strategy` with `k >= 1`,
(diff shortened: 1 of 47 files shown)


le1nux:
## Follow-up: real-data config, a broken config fixed, and a test that would have caught it

### `config_nemotron3_nano_30b_a3b_fsdp2.yaml` could never have started

Its `lr_scheduler` block used `num_warmup_steps` / `num_training_steps` / `min_lr`. The registered schema for `scheduler.linear_warmup_cosine_annealing_lr` requires `warmup_steps` / `total_steps` / `initial_lr` / `max_lr` / `final_lr`, so the component factory rejected it:

```
Invalid keys ['num_warmup_steps', 'num_training_steps', 'min_lr'] for config
`scheduler.linear_warmup_cosine_annealing_lr`
```

My fault, and my existing test did not catch it because `test_full_config_file_is_parseable` only checked that the YAML parsed — never that any component's config keys were valid. `tests/models/nemotron/test_shipped_config_schemas.py` now walks every component node in each shipped config and applies the factory's own key validation, plus a required-key check that reuses the factory's alias resolution so deprecated `AliasChoices` aliases (e.g. `wrapped_model` for `model_parts`) still count. I confirmed the checker rejects the exact key set that was in the 30B config.

It also caught a stale figure: the lorem-ipsum header's active-parameter column was shifted by one row (16-layer is 1.15B active, not 0.77B).

### New: `config_fineweb_nemotron_nano_fsdp2.yaml`

10.09B parameters / 1.57B active, sequence length 4096, vocab 128256, fused Mamba-2 kernels, `grouped_mm` experts, 131k tokens per step, 15,258 steps.

Train and eval data are byte-exact, **provably disjoint** slices of a document-shuffled FineWeb sample-10BT — train is tokens `[0, 2e9)`, eval is the last 2M. The cut is by byte offset, so disjointness needs no bookkeeping. This mattered: my earlier held-out slice was carved from the tail of the *full* training file, so pointing training at that file would have contaminated it.

### Verified end to end on the real data (4x A100-80GB, both exit 0)

| variant | steps | initial loss | final loss | held-out loss | peak memory |
|---|---|---|---|---|---|
| 16 layer (shipped) | 6 | 11.95 | 8.99 | 8.99 @ step 5 | 49.0 GiB |
| 9 layer | 40 | 11.94 | 6.96 | 7.38 @ step 20 | 33.8 GiB |

Two checks worth calling out. The initial loss matching `ln(128256) = 11.76` confirms the vocabulary matches the tokenization of the data. The held-out loss *tracking* the train loss at the same step is the check on the split — a noticeably lower eval loss would have indicated leakage.

Both runs used this config's architecture, data, optimizer and parallelism unchanged, differing only in run length (token budget capped, warmup scaled to match, checkpointing disabled to avoid a 113 GiB write). Neither is a convergence run.

**No throughput number is quoted, deliberately.** Both runs shared their GPUs with an unrelated job at 100% utilization, so their samples/s and MFU measure contention, not this configuration.

Note that item 3 of #460 (the pipeline stage packer) still applies here; once that lands, the `get_stages` override in `NemotronStagesGenerator` becomes redundant.

